PR #7, opened 2017-11-24
feat: update stripe pod

Atala:
@alexsegura can you tell me if this work with your xcode version?


alexsegura:
Sorry, I did the same in 48be2934c332d3b5e1b5542d9edafd6c70805f06

I was afraid that it wouldn't compile due to different XCode versions.

PR #30, opened 2018-02-16
Notify courier upon task assignation/unassignation

Atala:


alexsegura:
This doesn't work on iOS: websocket message is never received when the app is running in the backgound. This is documented [here](https://developer.apple.com/library/content/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/BackgroundExecution/BackgroundExecution.html). 

As a first implementation, on iOS let's open an [Alert](https://facebook.github.io/react-native/docs/alert.html) when the app is active to notify the user.

App.js
```diff
@@ -157,212 +153,279 @@ const routeConfigs = {
...
+  }
+
+  initializeRouter(baseURL, user) {
+    let client = null,
+      ws = null
+
+    if (baseURL) {
+      client = API.createClient(baseURL, user)
+      if (user && user.isAuthenticated() && (user.hasRole('ROLE_COURIER') || user.hasRole('ROLE_ADMIN'))) {
+        ws = this.createWebsocket(client)
```
alexsegura: The WebSocket is created only if the user is already authenticated when launching the app. 
The following use case doesn't work: 
- Logout
- Close app
- Open app
- Login
- Go to tasks screen

In this case, `ws` is `undefined`, and if you go to the tasks screen, it hangs. 
We need to send an event when the user logs in, to create the WebSocket.

src/page/courier/TasksPage.js
```diff
@@ -106,13 +107,13 @@ class TasksPage extends Component {
         { cancelable: false }
       )
     })
-
   }
 
   onWebSocketConnect() {
     this.props.navigation.setParams({ connected: true })
```
alexsegura: This is not sufficient: when entering this screen, as the WebSocket should already be connected, this method is never called, and the "wifi" icon is never green.

src/page/courier/TasksPage.js
```diff
@@ -70,25 +62,34 @@ class TasksPage extends Component {
...
       detailsModal: false,
     }
+
+    this.onMapReady = this.onMapReady.bind(this)
   }
 
   componentDidMount() {
     KeepAwake.activate()
+
+    Settings.addListener('websocket:connect', this.onWebSocketConnect.bind(this))
```
alexsegura: This listener seems useless, as the WebSocket connection is created earlier.

alexsegura:
I think using React Navigation to pass "global" variables was a mistake... my bad 😑

I've been trying, but I can't find a way to update the WebSocket (for example, to initialize it on login). 
I have tried to use [NavigationActions.setParams](https://reactnavigation.org/docs/navigation-actions.html#setparams) but I can't understand how to make it work...
I think we should use some kind of [registry pattern](https://martinfowler.com/eaaCatalog/registry.html) to store `baseURL`, `client`, `user`, `ws`

alexsegura:
I implemented registry pattern in 38b2128490d767d15ae81e60a04241cd1ec807e1

It works the same way. What do you think?
It looks way simpler.

src/page/courier/TaskListPage.js
```diff
@@ -15,6 +16,29 @@ class TaskListPage extends Component {
     }
   }
 
+  componentDidMount() {
+    Settings.addListener('websocket:message', this.onWebSocketMessage.bind(this))
```
alexsegura: You need to store `this.onWebSocketMessage.bind(this)` in a variable, or `removeListener` won't work as expected.

App.js
```diff
@@ -157,212 +154,271 @@ const routeConfigs = {
...
+
+    AppUser.load()
+      .then(user => {
+        Settings
+          .loadServer()
+          .then(baseURL => this.initializeRouter(baseURL, user))
+      })
+  }
+
+  onWebSocketMessage(event) {
```
alexsegura: Good idea to have bufferized the notifications. 
But I think it is wrong to manage it through the state: each time you change the state, the component re-renders, but nothing has actually changed. 
I think the buffer should be extracted to a dedicated class.

alexsegura:
@Atala I splitted this branch and rebased it on [websocket-global](https://github.com/coopcycle/coopcycle-app/tree/websocket-global). 

Relevant commits are bc0126a93bdbc8f01e48cfc8e4983c7677d00291 & 3f5425c14463443168b862a89507cc444c68c9eb

alexsegura:
The branch `websocket-global` has been merged on master 🙂

Atala:
React Navigation is a pain in the ass -_- I don't get how to do a global redirect with it. (i.e. a redirect from App.js)

No problem with the Registry pattern, it's just a replacement for a global redux store I guess.

couleurmate:
the `delay` is too short to be really noticed. The notification should disappear until user's action

alexsegura:
@couleurmate it was only a first implementation until we have remote push notifications, because what we have done doesn't work when the app is in the background. 

Work continues in #56.

PR #39, opened 2018-02-19
Date picker to filter tasks

Atala:


alexsegura:
I couldn't merge directly because it was not possible to interact with the map anymore.

I picked the header stuff but didn't pick the calendar stuff. 
The library `react-native-dates` doesn't manage translations correctly, and won't unless modified. 

You should avoid using too much absolute positioning / z-index, it is hard to manage.

alexsegura:
`react-native-dates` looks not really maintained. Is there something better out there?

Atala:
Agree about `zIndex`, I got confused/tired at the end - will finish properly.

Why `react-native-dates` won't manage translations properly?

alexsegura:
Because `react-native-dates` does not use `moment-with-locales` +  "Previous" and "Next" are hardcoded 😐

`react-native-calendars` looks cool 😎

I close this PR, let's continue in #43

PR #47, opened 2018-02-23
Add calendar picker on tasks pages

### LINKED ISSUE #3: Use Redux

### LINKED ISSUE #43: Add calendar picker on task list
A calendar should open when clicking on the date in `DateSelectHeader`, for example.

Atala:
Fix #43 
Fix #3

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,10 @@
     "react-native-modal-selector": "^0.0.24",
     "react-native-swipeout": "^2.3.3",
     "react-navigation": "^1.0.0",
+    "react-redux": "^5.0.7",
+    "redux": "^3.7.2",
+    "redux-logger": "^3.0.6",
+    "redux-thunk": "^2.2.0",
     "slugify": "^1.1.0",
     "tipsi-stripe": "4.4.0",
     "underscore": "^1.8.3"
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "react-navigation": "^1.0.0",
     "react-redux": "^5.0.7",
     "redux": "^3.7.2",
+    "redux-devtools-extension": "^2.13.2",
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.2.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -8,17 +8,24 @@
   },
   "dependencies": {
     "@mapbox/polyline": "^0.2.0",
+    "metro": "^0.26.0",
```
alexsegura: Is this needed? It should be required by `react-native`.

src/page/courier/TaskPage.js
```diff
@@ -243,7 +218,7 @@ class TaskPage extends Component {
 
     const swipeoutLeftButton = {
       component: this.renderSwipeoutLeftButton(),
```
alexsegura: Error when opening a task: `this.renderSwipeoutLeftButton is not a function`

alexsegura:
Redux ALL THE THINGS 🙂

Just one thought: the button to open the calendar is small and very close to the arrow, it is hard to click on it on a real device (see screenshot below when removing `transparent` property). 

I think it would be better to toggle the calendar by tapping on the date. 

capture d ecran 2018-02-24 a 03 10 07

package.json
Atala: Yes sorry, I'll remove it. I had issue with the `metro ` cache.

Atala:
Ok for this change. Do you think it is intuitive enough that one can click on the date to open the calendar? I think yes.

alexsegura:
I think yes ✅

Also, the arrow buttons should take all the remaining space available.

Atala:
@alexsegura I implemented all the changes you requested?

Can you check and merge if it's ok for you, so Elias can start the websocket stuff?

couleurmate:
Saw it late, but I think it's not fundamental to have a calendar picker per se, the arrows are enough. However, it would be more intuitive to click on the date, at least that what's done on most calendar app on Android.

Issue #21, opened 2018-02-08
Translate app


alexsegura:
- [ ] Switch all messages to English
- [ ] Add a setting to change language
- [ ] By default, set language to instance's default language (new endpoint needed)

eliascodes:
* [ ] change `moment` locale and google places language/region based on detected language

alexsegura:
Fixed in #48

PR #48, opened 2018-02-26
I18n

eliascodes:
Related to #21 

Hi everyone!

I've added some basic i18n scaffolding to the app, I've made the following assumptions, which I'm happy to change under advisement:
* There is no preference as to i18n library (I've used [react-i18next](https://react.i18next.com))
* Fallback translation language is set to English
* The translation resources are currently just organised into one namespace. If you don't want this, please advise what kind of organisation of translations you'd like (e.g. namespaced by view, or by feature, etc.)
* The translations are just straight translations of the existing hardcoded copy. I've not exploited any of the more advanced features like pluralisation or interpolation.
* Added a `LOCALE` config option to change `moment` and Google API locales.


App.js
```diff
@@ -192,11 +194,11 @@ export default class App extends Component {
       })
   }
 
-  onWebSocketMessage(event) {
+  onWebSocketMessage = (event) => {
```
Atala: You're doing thius change to have `this` automatically binded that's it? (ES6-syntax)

Not sure if it's ok to mix two styles for method declarations, maybe we should update all methods?

android/app/src/main/java/fr/coopcycle/MainApplication.java
```diff
@@ -26,6 +33,13 @@ public boolean getUseDeveloperSupport() {
     protected List<ReactPackage> getPackages() {
       return Arrays.<ReactPackage>asList(
           new MainReactPackage(),
+            new StripeReactPackage(),
```
Atala: Is it a `react-native` good practice to do this?

src/components/AddressTypeahead.js
```diff
@@ -80,14 +80,14 @@ export default class AddressTypeahead extends Component {
...
-          language: 'fr', // language of the results
+          language: AppConfig.LOCALE, // language of the results
           types: 'geocode', // default: 'geocode'
         }}
         styles={ styles }
         nearbyPlacesAPI="GoogleReverseGeocoding" // Which API to use: GoogleReverseGeocoding or GooglePlacesSearch
         GoogleReverseGeocodingQuery={{
           // available options for GoogleReverseGeocoding API : https://developers.google.com/maps/documentation/geocoding/intro
-          region: "fr"
+          region: AppConfig.LOCALE
```
Atala: I don't think this is correct. It will not work for Belgium for example (locale would be `fr` and `flamish`). I guess it will be better to have another setting `AppConfig.COUNTRY_NAME` (which will be somehow be related to the app store country in which the app is distributed)

What do you think?

src/i18n/locales/fr.json
```diff
@@ -0,0 +1,91 @@
...
+    "CONNECTED_TO": "Connecté à",
+    "CHANGE_SERVER": "Choisir un autre serveur",
+    "LOADING": "Chargement",
+    "HELLO": "Bonjour",
+    "DETAILS": "Informations personnelles",
+    "ADDRESSES": "Adresses",
+    "ORDERS": "Commandes",
+    "SIGN_OUT": "Déconnexion",
+    "CONTINUE": "Continuer",
+    "CHANGE_QUANT": "Modifiez la quantité",
```
Atala: `Modifier la quantité`

src/i18n/locales/fr.json
```diff
@@ -0,0 +1,91 @@
...
+    "ENTER_PAY_DETAILS": "Veuillez entrer vos coordonnées bancaires",
+    "PAY": "Payer",
+    "CONNECTING_SERVER": "Connexion au serveur",
+    "PLS_WAIT": "Veuillez patienter",
+    "WAITING_FOR_ORDER": "En attente d'une commande",
+    "LOADING_ORDER": "Chargement de la commande",
+    "CALCULATING_ROUTE": "Calcul de l'itinéraire",
+    "CALCULATING_YOUR_POS": "Calcul de votre position",
+    "TRACKING_YOUR_POS": "Suivi de votre position",
+    "CONN_LOST": "Connexion perdue, Reconnexion",
```
Atala: `Connexion perdue, reconnexion`

Atala:
@eliascodes seems good to me, except [this](https://github.com/coopcycle/coopcycle-app/pull/48/files#diff-845f80285e0cc674691c5b7cea34234aR90) - it should be a different setting I think.

Also could you please rebase your changes?

App.js
eliascodes: > You're doing thius change to have this automatically binded that's it? (ES6-syntax)

Yes, that's the only reason.

> Not sure if it's ok to mix two styles for method declarations, maybe we should update all methods?

The alternative is to bind `onWebSocketMessage` in the `constructor`. Whichever you prefer!

android/app/src/main/java/fr/coopcycle/MainApplication.java
eliascodes: I don't know, these files were automatically updated when I ran `link`/`start` 🤷‍♂️

App.js
Atala: Let's keep binding in the `constructor` for now, for consistence.

android/app/src/main/java/fr/coopcycle/MainApplication.java
Atala: Which version of `react-native` are you using?

eliascodes:
```
$ react-native --version
react-native-cli: 2.0.1
react-native: 0.53.2
```

alexsegura: Yep, there is a problem when running `react-native-link`, it duplicates some lines in the android project

eliascodes: OK, I can manually remove duplicated lines in the android source, unless you know of a better solution?

alexsegura: Yes please remove those lines. 
I need some time to understand the cause. I suspect the Gradle plugin because it affects only Android.

src/page/RestaurantPage.js
```diff
@@ -36,7 +37,7 @@ class RestaurantPage extends Component {
     const { cart } = this.state
     cart.addMenuItem(menuItem)
 
-    this.cartFooter.animate()
+    this.cartFooter.getWrappedInstance().animate()
```
eliascodes: Note that this is because `CartFooter` is wrapped in a HOC using `translate`. See c6fec3d.

This is a simple solution, but I don't know if it's the best one, since it relies on `RestaurantPage` understanding that `CartFooter` is wrapped. Ultimately I think this pattern of calling a method on a child component in the parent is breaking encapsulation. Ideally the `animate` method would be called by `CartFooter` triggered by a change to props or something.

Anyway, I'm happy to leave it like this, I just wanted to draw your attention to it.

Also, if you know of anywhere else in the codebase where this pattern is used, please let me know, and I will double check this whether this HOC-wrapping bug is present. Thanks. 😃

src/AppConfig.json.dist
```diff
@@ -1,4 +1,6 @@
 {
     "GOOGLE_API_KEY": "",
-    "STRIPE_PUBLISHABLE_KEY": ""
-}
\ No newline at end of file
+    "STRIPE_PUBLISHABLE_KEY": "",
+    "LOCALE": "",
```
Atala: Can you just add default values here?

LOCALE = "fr"
COUNTRY_NAME = "fr"

Afterwards I am ok to merge

alexsegura:
@eliascodes thanks a lot!!! 💯
@Atala I merged it now because I think @eliascodes has addressed all your comments and I need to add a new screen, so I wanted to avoid more rebases.

alexsegura:
@eliascodes I removed `react-native-locale-detector` because I was not able to compile the Android project for production. 

capture d ecran 2018-03-05 a 14 44 13

Actually the repo is quite old and includes directly old versions of `react-native-i18n`

capture d ecran 2018-03-05 a 14 43 05

I fixed it in 3471972f5f445bc2653afb707c7bb4436cb7a061 by using directly the latest `react-native-i18n`, which returns the locale as well. 

Now let's release a new version 🙂

eliascodes:
@alexsegura I'm sorry that slipped by, I should have checked before submitting the PR.

Perhaps we can add two more steps to the travis build config in #57 so we ensure all PRs successfully compile.

PR #56, opened 2018-03-01
Remote push notifications

alexsegura:
Let's introduce remote push notifications _for real_. 

It means there is a also a server side part to store the tokens associated to each device, and to send payloads to push servers from Apple & Google. I started a [branch](https://github.com/coopcycle/coopcycle-web/commits/remote-push-notifications). 

@Atala I'm still trying to understand what is the best option to share certificates, I'll tell you shortly. 
You need to do it for Google because it only works on real devices.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,7 @@ buck-out/
 # Pods/
 
 src/AppConfig.json
+/android/app/google-services.json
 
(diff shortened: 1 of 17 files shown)


Atala:
@alexsegura which help do you need?

Issue #66, opened 2018-03-11
Calls to `translate` wrapped components causes runtime error


eliascodes:
I was doing a bit of informal testing and spotted a couple more instances of the bug I identified in #48 (also see c6fec3d).

I'll submit a fix shortly, and do a project-wide search for uses of `ref` to double check for other occurrences.

PR #67, opened 2018-03-11
i18n [bug]: fix more HOC related bugs

eliascodes:
Fixes https://github.com/coopcycle/coopcycle-app/issues/66

This fixes the same bug identified in c6fec3d, where wrapping components
in a HOC causes the pattern of calling methods directly on component
refs to break.

For now, this is fixed in the same way as in c6fec3d

I've searched every occurrence of `ref` in the project and I don' think there are any more places where this bug could surface.

Also removes a couple of unnecessary lines.

### DIFF
--- a/src/components/DateTimePicker.js
+++ b/src/components/DateTimePicker.js
@@ -54,4 +54,4 @@ class DateTimePicker extends Component {
   }
 }
 
-export default translate()(DateTimePicker)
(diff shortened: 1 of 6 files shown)


Atala:
Thanks for the fix, merging.

Also note that I have added tests in the application so feel free to add some :) (although tests related to i18n seems kind of painful to manage)

Issue #37, opened 2018-02-19
Quickfix : enable the use of the app when socket is disconnected


Atala:
Instead of putting the "Reconnection.." loader for websocket above the map, just put a small loader. For example in replacement of the wifi icon on the above.

The task list will be updated at reconnection anyway.

@alexsegura what do you think?

alexsegura:
Yep 🙂
Actually I did this because right now `WebSocketClient` doesn't manage any kind of "buffer" when it is disconnected. 
I would say that a great improvement would be to manage disconnections correctly: 
- When the WebSocket is disconnected, calls to `WebSocketClient.send` should be stored _somewhere_ (for example, in `LocalStorage`). 
- When the WebSocket reconnects, it should send all the waiting messages.

Atala:
+ an (orange?) horizontal bar should alert the user that no automatic refresh should be triggered.

Do you think we should handle the WS lifecycle inside Redux? IMHO it would be nicer, because it will clean all the `loading..` mess we have right now.

Maybe a nice task for Elias?

eliascodes:
@Atala Hi!

I'm happy to take a look at this later this week/weekend.

I'd like a bit of clarity on your last comment though:

> Do you think we should handle the WS lifecycle inside Redux

Redux currently isn't being used in the project, are you intending to start using it? (I noticed #3, but it has no detail 😃). If so, that is quite a lot of work compared to this issue, and it might be worth handling that separately. What do you think?

(Although in principle I agree, we can handle all the websocket stuff with a redux middleware)

Atala:
Actually I added Redux in #47 , the PR should be merged tonight, so you can start from this branch. Just a little part of the app state is in Redux yet, but we should migrate the whole thing step-by-step.

I think it's nice if you take on this issue, as you will have to see some part of platform as well (for testing purposes).

On my side, I will be happy to see the code for a redux-middleware, never seen that in my life ;)

Feel free to ask any questions about the specs or the code, as it is not very detailed yet.

eliascodes:
> On my side, I will be happy to see the code for a redux-middleware, never seen that in my life ;)

The idea is if you can isolate WS code inside a middleware, the rest of your application can interact with it entirely through actions, which should in theory make things more declarative, and testable.

So in application code, we might dispatch actions like:
```js
dispatch({ type: 'WS_CONNECT', payload: 'ws://domain.com/foo' })

// or

[...]

PR #72, opened 2018-03-16
[WIP] Websocket management

eliascodes:
**WIP - DO NOT MERGE**

Related to #37 

Hi all!

Here's where I'm at right now. I've taken some liberties with the organisation of the redux code. I hope you don't mind, but if you do, I'm happy to put it back the way it was. I've explained my reasoning in the commit message for 62727ad.

I decided it was probably better to keep the middleware relatively simple and keep most of the logic in the `WebSocketClient`, so I added the message queuing there rather than in the middleware.

I'm happy to discuss the direction I'm going in here if you have any concerns!

I will be able to get back to this on Sunday probably. :)

#### TODO:
* [ ] More manual testing
* [ ] Write a few more automated tests
* [ ] Decide what to do if `Registry.initWebSocketClient` rejects


src/page/courier/TaskListPage.js
```diff
@@ -8,9 +8,12 @@ import moment from 'moment/min/moment-with-locales'
 import TaskList from '../../components/TaskList'
 import DateSelectHeader from '../../components/DateSelectHeader'
 import { whiteColor } from '../../styles/common'
-import { changedTasks, loadTasksRequest } from "../../store/actions"
 import { translate } from 'react-i18next'
 import { localeDetector } from '../../i18n'
+import {
+  loadTasks, markTaskDone, markTaskFailed,
+  selectTasksList, selectTaskSelectedDate, selectIsTasksLoading,
+} from '../../redux/Tasks'
```
Atala: Hey,

I like the idea of splitting out things, but I think a better namespacing would be `Courier`, `Customer` and `Restaurant` - so it will stick to the  3 different user kinds.

App.js
```diff
@@ -219,20 +221,34 @@ class App extends Component {
       user: null,
       serverError: false,
     }
-    this.onWebSocketMessage = this.onWebSocketMessage.bind(this)
+
+    store.subscribe(() => {
```
Atala: Nice - we will be able to deal with remote notifications the same way, did you check [this](https://github.com/reactjs/redux/issues/303#issuecomment-125184409) ?

src/page/courier/TasksPage.js
```diff
@@ -81,29 +86,12 @@ class TasksPage extends Component {
       }
     })
 
-    this.onConnectHandler = this.onWebSocketConnect.bind(this)
```
Atala: Nice :)

src/redux/App/reducers.js
```diff
@@ -0,0 +1,30 @@
+/*
+ * App reducer, dealint with non-domain specific state
```
Atala: Typo

src/redux/Tasks/__tests__/actions.test.js
```diff
@@ -0,0 +1,209 @@
...
+      actionType: DONT_TRIGGER_TASKS_NOTIFICATION,
+    },
+  ]
+    .forEach(({ actionCreator, actionType }) => {
+      test(`${actionType}`, () => {
+        expect(actionCreator()).toEqual({ type: actionType })
+      })
+    })
+
+  test('loadTasks | Successful request', () => {
```
Atala: Thumbs all the way up

src/redux/middlewares/WebSocketMiddleware/index.js
```diff
@@ -0,0 +1,122 @@
...
+  return (next) => (action) => {
+    if (! isWsAction(action)) {
+      return next(action)
+    }
+
+    validateAction(action)
+
+    switch (action.type) {
+      case INIT:
+        if (! isInitialised) {
```
Atala: typo `initialised`

src/websocket/WebSocketClient.js
```diff
@@ -88,9 +103,11 @@ class WebSocketClient {
     this.openCount = this.openCount + 1
     this.closeCount = 0
 
-    this.options.onConnect()
+    this.options.onConnect(event)
     if (this.openCount > 1) {
-      this.options.onReconnect()
+      this.options.onReconnect(event)
+
+      JSONAsyncStorage.consume('@WsMsgQueue', (msg) => this.send(msg))
```
Atala: I don't think this is the correct place to call this. Will it be called if you don't connect/disconnect at least once?

Atala:
Great job! I just added some comments here and there.

Maybe we could cherry-pick some commits (i18N fixes, coverage support, tests for tasks fetching) so we reduce the size of the PR and see things more clearly?

src/redux/Tasks/reducers.js
```diff
@@ -0,0 +1,172 @@
...
+moment.locale(localeDetector())
+
+
+/*
+ * Intital state shape for the task entity reducer
+ */
+const tasksEntityInitialState = {
+  fetchError: null,                // Error object describing the error
+  isFetching: false,               // Flag indicating active HTTP request
+  isStale: false,                  // Flag indicating state fragment should be re-fetched
```
Atala: Nice

src/websocket/WebSocketClient.js
eliascodes: The first time `open` is emitted, `@WsMsgQueue` is empty, and `consume` will not be called.

Would you prefer `consume` to be called every time `open` is emitted?

src/redux/Tasks/reducers.js
eliascodes: Yes, it's not used yet. I couldn't quite figure out what the best condition would be under which to mark tasks as stale. Maybe I should just remove it, it's easy enough to add it back if we decide we need it.

What do you think?

eliascodes:
@Atala thanks for your comments! I'll get onto them soon.

> Maybe we could cherry-pick some commits (i18N fixes, coverage support, tests for tasks fetching) so we reduce the size of the PR and see things more clearly?

No problem, I'll stick them in a separate PR.

On another note, the android build seems to be failing, and while I can reproduce it locally, I'm a little stuck on what the problem might be (I've not much experience debugging Java). There seems to be a null pointer exception without an error message, and the stacktrace points to `:app:processDebugManifest`, but googling hasn't enlightened me much on why it might be failing. Have you had this error before?

App.js
eliascodes: Yeah, if you like, I can add a little `observe` utility like the one Dan illustrates in that comment and use that instead of subscribing directly?

src/page/courier/TaskListPage.js
eliascodes: Are you thinking that this namespace would just be reflected in the directory structure, or also the shape of the state?

I think the former is fine, but less sure about the latter.

eliascodes:
Also I noticed that we call the `WebSocket` constructor with three arguments, where the third sets the authorization header.

https://github.com/coopcycle/coopcycle-app/blob/341ab26458b16961a083b2cedb9cd0d71feebde9/src/WebSocketClient.js#L50-L54

But this isn't in the spec for the [Websocket API](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket). I think the third argument is just ignored. Looking at [stackoverflow](https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api), there doesn't seem to be a way to add custom headers to the websocket connection. From the comments in that answer, seems like an alternative is to send the token either as a query param or as the first ws message. Should I open a separate issue about this?

src/redux/middlewares/WebSocketMiddleware/index.js
eliascodes: This is not a typo, I usually prefix boolean variables with `is` or something similar so that it can be read as a question: "is initialised? yes(true) or no(false)".

Happy to change it to just `initialised` if you prefer though.

src/websocket/WebSocketClient.js
Atala: In my mind a nice use-case, would have been not to block the interface (with a big loader) when the WS connection is not open. Before opening we don't block the user to do things so we save the message right away.
Is it clear? What do you think?

src/redux/Tasks/reducers.js
Atala: I am not sure you need this boolean, as you can check `fetchError === null` (or sthg equivalent) to know if you failed to fetch.
In any case please remove it if it is not used.

Atala:
I've had build issues as well (same one), will try to find a solution (not a Java expert either)

src/websocket/WebSocketClient.js
eliascodes: If I understand correctly, you would prefer never to have to show any kind of "loading" spinner due to the websocket not yet being connected.

I can just move the call to `consume` out of the `if` statement, so that it will be called on every `open` event.

App.js
Atala: yeah Dan says it's nicer so let's follow his advice. Maybe in a `notifications` file/folder ?

src/redux/Courier/taskSelectors.js
```diff
@@ -0,0 +1,20 @@
...
+ * Here we use `reselect`, which allows us to memoize computed property values,
+ * benefitting performance.
+ */
+import { createSelector } from 'reselect'
+
+export const selectTaskSelectedDate = state => state.ui.tasks.selectedDate
+export const selectTriggerTasksNotification = state => state.entities.tasks.triggerTasksNotification
+export const selectIsTasksLoading = state => state.entities.tasks.isFetching
+export const selectIsTasksLoadingFailure = state => state.entities.tasks.fetchError
+export const selectTasks = state => state.entities.tasks
```
Atala: Should be `state.entities.tasks.items` no ? Otherwise `selectTasksList` won't work

Issue #76, opened 2018-04-04
Add a signup screen in the app


Atala:
Some inspiration can be found [here](https://github.com/coopcycle/coopcycle-js/blob/master/src/components/RegisterForm.js) (for the logical part)

alexsegura:
Fixed in #83

PR #83, opened 2018-04-12
Register page

eliascodes:
Related to #76 

Hey everyone, only managed to get around to this today. There are a couple of missing pieces, related to my questions below. I've setup a simple register form, screenshots also below. Let me know if you have UX suggestions, and feel free to correct my (probably) terrible French translations 😅 .

Some things I was wondering about:
* Is the API documented anywhere?
* What error codes can be expected from the `/api/register` endpoint? 400, 409, 500, 502 ...? I haven't currently got any good error handling on the register request because I don't know what to expect.
* Would you like a successful registration to do anything different from a successful login? Currently I'm just passing the same `onLoginSuccess` and `onLoginFailure` callbacks to both the `LoginForm` and the `RegisterForm`.
* I tried to make these changes as simply as possible, which means no integration with the redux store. If that's something you want, let me know.
* Do you want any kind of client-side validation, or are you happy to leave it all to the sever for now?
* This will only work for customer sign-ups. Do you also want couriers/restaurants/etc to be able to sign up?
* I noticed that when I submitted my phone number (of the form 07xxxxxxxxx), the leading zero was removed (which suggests it's converted to an integer by the API). This seems like a problem. 

Screenshots:
screen shot 2018-04-12 at 16 25 55
screen shot 2018-04-12 at 16 26 14
screen shot 2018-04-12 at 16 26 25

Thanks 👍


eliascodes:
> Register endpoint repsonds with 404 (easier to trigger : "email already exists") or 200.

These two lines suggest that the endpoint returns a `400`, not a `404` in case of an error, unless I'm mistaken?:

https://github.com/coopcycle/coopcycle-web/blob/668db0ea6ceb10a609e43b1585221982ae846639/src/AppBundle/Action/Register.php#L93

https://github.com/coopcycle/coopcycle-web/blob/668db0ea6ceb10a609e43b1585221982ae846639/src/AppBundle/Action/Register.php#L106

Confirmed by manual testing:
screen shot 2018-04-19 at 18 16 15


> Maybe it is easy to find a lib that define some input type as email, phone number, aso?

I quite like [`joi`](https://www.npmjs.com/package/joi) for validation, though it needs a plugin to support phone numbers (something like [joi-phone-number](https://www.npmjs.com/package/joi-phone-number)). Neither of the main alternatives ([`validate.js`](https://www.npmjs.com/package/validate.js) and [`validator`](https://www.npmjs.com/package/validator)) support phone-numbers out of the box either.

Atala:
Yes my bad, obviously it should & does return a `400` (!)

L. 94 does indeed return a correct JSON response.
Actually if the validation was done correctly L. 106 shouldn't be reached. I think in your example you met an exception. Do you have the payload or more information about the error?

`joi` seems cool let's go for it!

Atala:
I think in case we reach L.106 on server side we should have : 
 * an error log so back-end devs can investigate
 * a very general and unspecific error message on the app (no `responseText` -> `An error occured` or something of the kind)

eliascodes:
Oops, looks like `joi` doesn't work with react-native because one of its dependencies is incompatible (`Module `assert` does not exist in the Haste module map`). I'll use one of the alternatives. If you have a preference, let me know.

src/components/RegisterForm.js
```diff
@@ -0,0 +1,221 @@
...
+  static schema = {
+    givenName: {
+      presence: { message: i18n.t('INVALID_GIVEN_NAME') },
+      length: {
+        minimum: 1,
+        message: i18n.t('INVALID_GIVEN_NAME'),
+      },
+    },
+    familyName: {
+      presence: { message: i18n.t('INVALID_GIVEN_NAME') },
```
alexsegura: Bad error message here (more below)

src/i18n/locales/fr.json
```diff
@@ -94,6 +94,20 @@
     "LAST_TASK_EVENT": "Dernier évènement {{fromNow}}",
     "HISTORY": "Historique",
     "SETTINGS": "Réglages",
-    "SETTING_KEEP_AWAKE": "Désactiver la mise en veille"
+    "SETTING_KEEP_AWAKE": "Désactiver la mise en veille",
+    "OR_REGISTER": "Vous n'avez pas de compte? Registre",
```
alexsegura: Let's use `Vous n'avez pas de compte ?` (there is a space in french before `?`)

src/i18n/locales/fr.json
```diff
@@ -94,6 +94,20 @@
...
+    "OR_LOGIN": "Vous avez déjà un compte? S'identifier",
```
alexsegura: Let's use `Vous avez déjà un compte ?` (there is a space in french before `?`)

alexsegura:
LGTM

Can you fix the bad error messages please?

src/components/RegisterForm.js
```diff
@@ -0,0 +1,221 @@
...
+
+  render() {
+    return (
+      <View>
+        <Form>
+          <Item stackedLabel>
+            <Label>{this.props.t('GIVEN_NAME')}</Label>
+            <Input
+              autoCorrect={false}
+              autoCapitalise="none"
```
alexsegura: Should be `autoCapitalize`

alexsegura:
Thank you 💯
This will be included in the next release `0.7.3` 😊

PR #120, opened 2018-09-07
Refactor checkout

### LINKED ISSUE #117: Refactor checkout using React Navigation + Redux
The checkout process was made a long time ago, at uses a lot of "hacks".
Let's refactor it with experience 🤓

- [ ] Use React Navigation to open modals
- [ ] Allow choosing product options
- [ ] Store the cart in Redux
- [ ] Upgrade tipsi-stripe

alexsegura:
Fixes #117 

- [x] Use Redux to store the cart state during checkout
- [x] Open modal for products with options
- [x] Use React Navigation for modals
- [x] Clear Redux state when checkout finished
- [x] Add options to order payload

⚠️ This Pull Request does not change the screens, which could be still improved, but at least it will be easier.

app_checkout_redux


src/navigation/checkout/ProductOptions.js
```diff
@@ -0,0 +1,139 @@
...
+    this.props.navigation.goBack()
+  }
+
+  renderFooter() {
+    if (this._enableAddToCartButton()) {
+      return (
+        <Footer>
+          <FooterTab>
+            <Button full onPress={ () => this._onPressAddToCart() }>
+              <Text>Ajouter au panier</Text>
```
eliascodes: Should this be internationalised?

src/Cart.js
```diff
@@ -82,15 +97,36 @@ class Cart {
     return _.reduce(this.items, function(memo, item) { return memo + item.quantity; }, 0);
   }
   clone() {
-    return new Cart(this.restaurant, this.items.slice());
+    return new Cart(this.restaurant, this.items.slice(0));
```
eliascodes: The `slice` only performs a shallow clone of the array, not of any objects within it. That is to say:
```js
var x = this.items.slice();

x === this.items; // false

x[0] === this.items[0]; // true
```
I wasn't sure whether this was intended behaviour or not, so I thought I should mention it.

Also `slice()` and `slice(0)` do the same thing.

alexsegura: Yep, I used `slice(0)` because I saw it somewhere else 😅

This `Cart` class does not really play well with Redux, I struggled for some changes to be detected. I think we may even remove it at some point, and store the items directly in the Redux store, but I wanted to make a quick refactoring.

src/navigation/CreditCardPage.js
```diff
@@ -27,34 +31,48 @@ class CreditCardPage extends Component {
...
   _onClick() {
 
-    const { cart } = this.props.navigation.state.params
+    const { cart } = this.props
 
     if (this.state.valid) {
+
       this.setState({ loading: true });
+
       Stripe.createTokenWithCard(this.state.params)
```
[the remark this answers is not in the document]
alexsegura: Ok, good catch. Let's do this on the next iteration.

src/Cart.js
eliascodes: Yeah I think ideally the store would just contain data types that are supported by JSON (so no class instances). The benefits of this are: (1) the state tree itself contains no complex objects with hidden state, so it can be cloned reliably, and (2) it can be serialised to a JSON string, which helps with things like server-side-rendering (not relevant in this case).

But as you say, for now it's just more important that things work 😄.

PR #177, opened 2019-04-07
Update tasks map, list and DateSelectHeader layouts Fixes #175

### LINKED ISSUE #175: Can't interact with calendar on Android
On Android, interacting with the calendar in the task list screen doesn't work. 
Maybe we should use the same technique as in [`RestaurantSearch`](https://github.com/coopcycle/coopcycle-app/blob/f2958ab9d5922b11e3968c0f2e70016ac7566428/src/components/RestaurantSearch.js) with absolute positioning + z-index

Screenshot_20190402-113538

vova1vova:
Fixes #175


alexsegura:
Thank you @vova1vova 🙂

PR #187, opened 2019-05-01
Use a different package name for debug android build

vova1vova:
Added a different package name for android debug build, so it's possible to have debug and production versions installed on the same device. I also made the app icon black and white for a debug build


vova1vova:
to fix the build, google-services.json (from Firebase) has to be updated (probably on travis). It has to contain both `fr.coopcycle.debug` and `fr.coopcycle` package names

PR #190, opened 2019-05-11
Use <Content /> Fixes #163

### LINKED ISSUE #163: Use KeyboardAvoidingView
On each screen where there is a submit button / form, we should use `KeyboardAvoidingView` to make sure the button stays clickable when the keyboard is visible. 

- [x] Add to credit card screen f2958ab9d5922b11e3968c0f2e70016ac7566428
- [ ] Add to login/register screen

https://facebook.github.io/react-native/docs/keyboardavoidingview

vova1vova:
Refactor `Credit card` screen to use `<Content />`. 
Check that `login/register` screen and `configure server` screen are using `<Content />` too, so that keyboard would be handled by `<Content />` component.

Update `login/register` screen to display `Next/Done` keyboard button and handle it


alexsegura:
Sorry for taking so much time 😔

Actually, I cherry-picked your commits directly in the master branch. 
I didn't cherry-pick f499fe1573f7f5e6d8431f27c34b1432979a11c4, because I think it belongs to a clean bump of React Native. 

Nice catch with the `extraScrollHeight` attribute!
Let's keep #163 open, because I think there are other screens which would need this kind of UX optimization. 

Thank you so much! 😍

PR #199, opened 2019-07-05
Bump target sdk version to API level 28

vova1vova:
Fixes https://github.com/coopcycle/coopcycle-app/issues/196

Bumped target sdk version to 28, as it's supported by React Native since version 0.59 https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#059 

I'm not sure if there is anything else that needs to be done in scope of this issue. There is a list of changes for API level 28 https://developer.android.com/about/versions/pie/android-9.0-changes-28 may be there is anything from it that is applicable to the app?

### DIFF
--- a/android/app/src/debug/AndroidManifest.xml
+++ b/android/app/src/debug/AndroidManifest.xml
@@ -1,0 +1,4 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <application
+        android:usesCleartextTraffic="true"/>
+</manifest>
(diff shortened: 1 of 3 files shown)


vova1vova:
Yes, for some reason it stopped working when I updated API level. And it was the same error that is mentioned here https://github.com/facebook/react-native/issues/24529 One of the solutions mentioned in this thread is to replace `network_security_config.xml` with `android:usesCleartextTraffic` https://github.com/facebook/react-native/issues/24529#issuecomment-507776241 This is what I did and it helped. As it's done only for `android/app/src/debug` build variant it should have an effect only on **debug** build

alexsegura:
Thank you @vova1vova 😊

PR #201, opened 2019-07-16
Provide 64 bits version for Android

### LINKED ISSUE #193: Provide 64 bits version for Android
The following message now appears on Google Play

Capture d’écran 2019-05-19 à 10 50 29

- https://android-developers.googleblog.com/2019/01/get-your-apps-ready-for-64-bit.html
- https://medium.com/@andriidrozdov/reactnative-and-android-64-bit-new-google-play-market-rules-what-to-do-584b067d6f1a
- https://developer.android.com/distribute/best-practices/develop/64-bit

vova1vova:
Fixes #193 

One consequence of this, is that the app becomes +13Mb bigger, due to 64-bit binaries (it would be 37Mb in total). If it becomes a problem, it's possible to start using existing setup for generating multiple APKs or add support for App Bundles, but that would work only for Google Play https://blog.swmansion.com/make-your-react-native-app-3x-smaller-44c993eda2c9

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -110,7 +110,7 @@ android {
         versionCode 49
         versionName "0.15.0"
         ndk {
-            abiFilters "armeabi-v7a", "x86"
+            abiFilters "armeabi-v7a", "x86", "arm64-v8a", "x86_64"
         }
         manifestPlaceholders = [
             tipsiStripeRedirectScheme: "coopcycle",


alexsegura:
Thank you so much @vova1vova ! 😃
At the moment, I just want to see those warnings on Google Play disappear haha. 
We'll see later if we can optimize the size

alexsegura:
[Here](https://facebook.github.io/react-native/docs/hermes/) they say that using Facebook's dedicated JavaScript engine would reduce app size. 

> For many apps, simply enabling Hermes will result in improved start-up time, decreased memory usage, and smaller app size.

Maybe it's too early to switch, but it can be a way to reduce app size.

PR #319, opened 2019-12-03
Use custom icons to distinguish pickup & dropoff markers

vova1vova:
Fixes https://github.com/coopcycle/coopcycle-app/issues/245

Added different icons for pickup and dropoff tasks. 

Use (almost) the same colors for the tasks in the list and on the map

Markers are darkgrey by default:
Simulator Screen Shot - iPhone X - 2019-12-03 at 20 50 08

Or use a tag color:
Simulator Screen Shot - iPhone X - 2019-12-03 at 20 47 49

### DIFF
--- a/src/components/TasksMapView.js
+++ b/src/components/TasksMapView.js
@@ -5,7 +5,7 @@ import {
   StyleSheet,
   Text,
   TouchableOpacity,
-  View
+  View,
 } from 'react-native'
 import {Callout, Marker} from 'react-native-maps'
 import ClusteredMapView from 'react-native-maps-super-cluster'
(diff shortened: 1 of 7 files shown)


src/navigation/task/styles/common.js
```diff
@@ -0,0 +1,87 @@
...
+    transform: [
+      {rotate: '45deg'},
+    ],
+  }
+}
+
+const markerIconStyle = task => {
+  return {
+    position: 'absolute',
+    fontSize: 24,
```
alexsegura: Maybe 24 is a little too much, 16-18 looks better. But it's a matter of taste 🙂

src/components/TasksMapView.js
```diff
@@ -191,9 +206,12 @@ class TasksMapView extends Component {
         identifier={ task['@id'] }
         key={ task['@id'] }
         coordinate={ task.address.geo }
-        pinColor={ pinColor(task) }
         flat={ true }
         ref={ this.markers.get(task['@id']) }>
+        <View style={ markerContainer }>
+          <View style={ markerBackgroundStyle(task) } />
+          <Icon type="FontAwesome" name={ markerIconName(task) } style={ markerIconStyle(task) } />
```
alexsegura: To avoid repeating, it's possible to refactor the icon into a dedicated React Component, which contains the icon/colors logic. With props, we could configure if we want to show the type icon, or the state icon.

Something like this:
```jsx
<TaskIcon task={ task } /> {/* cube/arrow-down */}
<TaskIcon task={ task } type="type" /> {/* cube/arrow-down */}
<TaskIcon task={ task } type="state" /> {/* check/remove */}
```

Issue #189, opened 2019-05-11
bell's ring


damienlapoitavelo:
I don't know why but when I open coopcycle app after attributing to me a delivery, a ring appears and ring hardly. The volume is so strong and unpleasant. Everybody in restaurant look at me. 
i'm not sure it's necessary to prevent me for a delivery when I open the app to see the deliveries.
Au pire Mex !!!!! tu me vires ce bruit strident et tu me le remplaces par quelque chose de plus discret parce que là la honte que j'me tappe à chaque fois !!!!!

PR #432, opened 2020-02-05
add an option to enable/disable 'tasks changed' alert sound

vova1vova:
Screenshot_1580926668

Fixes https://github.com/coopcycle/coopcycle-app/issues/189


src/redux/reducers.js
```diff
@@ -56,23 +56,27 @@ const tasksUiPersistConfig = {
           AsyncStorage.multiGet([
             '@Settings.keepAwake',
             '@Preferences.signatureScreenFirst',
-            '@Preferences.tasksFilters'
+            '@Preferences.tasksFilters',
+            '@Settings.tasksChangedAlertSound',
```
alexsegura: You can actually remove this. This code is here to "migrate" from the legacy way of storing data directly with `AsyncStorage`, to the new way with `redux-persist`. As this key is new, it won't be found in 100% use cases. 
You can just keep the default value for `tasksChangedAlertSound` in `resolve()`.

src/redux/reducers.js
```diff
@@ -82,6 +86,7 @@ const tasksUiPersistConfig = {
             keepAwake: false,
             signatureScreenFirst: false,
             excludeFilters: [],
+            tasksChangedAlertSound: false,
```
alexsegura: Default value should be `true` for backwards compatibility

src/redux/Courier/taskUiReducer.js
```diff
@@ -19,6 +19,7 @@ import {
 const tasksUiInitialState = {
   selectedDate: moment(), // Date selected by the user
   excludeFilters: [],     // Key-value pairs of active filters (e.g. status: 'done')
+  tasksChangedAlertSound: false,
```
alexsegura: Default value should be `true` for backwards compatibility

src/redux/reducers.js
```diff
@@ -56,23 +56,27 @@ const tasksUiPersistConfig = {
...
 
             let keepAwakeValue = keepAwake[1] ? JSON.parse(keepAwake[1]) : false
             let signatureScreenFirstValue = signatureScreenFirst[1] ? JSON.parse(signatureScreenFirst[1]) : false
             let tasksFiltersValue = tasksFilters[1] ? JSON.parse(tasksFilters[1]) : []
+            let tasksChangedAlertSoundValue = tasksChangedAlertSound[1] ? JSON.parse(tasksChangedAlertSound[1]) : false
 
             AsyncStorage.removeItem('@Settings.keepAwake')
             AsyncStorage.removeItem('@Preferences.signatureScreenFirst')
             AsyncStorage.removeItem('@Preferences.tasksFilters')
+            AsyncStorage.removeItem('@Settings.tasksChangedAlertSound')
```
alexsegura: You can remove this (see comment above)

src/redux/reducers.js
```diff
@@ -56,23 +56,27 @@ const tasksUiPersistConfig = {
...
+            '@Settings.tasksChangedAlertSound',
           ]).then(values => {
 
-            let [ keepAwake, signatureScreenFirst, tasksFilters ] = values
+            let [ keepAwake, signatureScreenFirst, tasksFilters, tasksChangedAlertSound ] = values
 
             let keepAwakeValue = keepAwake[1] ? JSON.parse(keepAwake[1]) : false
             let signatureScreenFirstValue = signatureScreenFirst[1] ? JSON.parse(signatureScreenFirst[1]) : false
             let tasksFiltersValue = tasksFilters[1] ? JSON.parse(tasksFilters[1]) : []
+            let tasksChangedAlertSoundValue = tasksChangedAlertSound[1] ? JSON.parse(tasksChangedAlertSound[1]) : false
```
alexsegura: You can remove this (see comment above)

alexsegura:
Thank you @vova1vova 🔉

PR #530, opened 2020-03-26
android: open app when restaurant receives new order

vova1vova:
Depends on https://github.com/coopcycle/coopcycle-web/pull/1256 Don't release until it's merged.

When https://github.com/coopcycle/coopcycle-web/pull/1256 is released and the backend sends both 'notification+data' and 'data-only' messages, the app would:

- in the foreground: handle only 'data-only' messages
- in the background: display 'notification+data' messages as notifications in the notification bar (firebase sdk handles this part - I think this part complicated the process a bit, and it would be easier if android app does it on the app side); when 'order:created' 'data-only' message is received - open the app (should work on devices up to android 10, on android 10 use 'notification+data' message as a fallback)


alexsegura:
Thank you @vova1vova

PR #688, opened 2020-07-19
Update README.md directions

jasonprado:
I was confused by a few steps of the initial build process so I clarified them a bit.


alexsegura:
Thank you @jasonprado

PR #745, opened 2020-09-01
fix: patch security provider, to enable TLS 1.2 for older Android devices (fixes #744)

### LINKED ISSUE #744: Android 4.4 users cannot connect to CoopCycle servers
**Describe the bug**
Users of Android version 4.4 and below see an error "Network request failed" when connecting to demo.coopcycle.org and some other servers (e.g. southwark.coopcycle.org).

Note: I have a pull request with a potential fix which I will link for review.

**To Reproduce**
1. Using an Android 4.4.2 device (e.g. an emulated 4.4.2 device in Android Studio),
2. Open the CoopCycle app for the first time,
3. Press "Choose a city",
4. Press the three-dot menu in the top right,
5. Type demo.coopcycle.org,
6. Press Submit.

The user sees "Network request failed" and is unable to continue.

**Expected behavior**
The list of Restaurants screen loads.

**Smartphone (please complete the following information):**
 - OS: Android 4.4.2
 - CoopCycle app

**Additional context**
Inspecting the TLS certificate for demo.coopcycle.org (see results [here](https://www.ssllabs.com/ssltest/analyze.html?d=demo.coopcycle.org&hideResults=on&latest)), it requires TLS 1.2 or higher (TLS 1.1 and below are marked as "No").
On Android 4.4 and lower, it appears to be a known problem that requests won't use TLS 1.2 by default. There is information [here](https://medium.com/tech-quizlet/working-with-tls-1-2-on-android-4-4-and-lower-f4f5205629a) about resolving the problem, and specific [docs](https://developer.android.com/reference/javax/net/ssl/SSLSocket#protocols) showing that TLS 1.2 is available from Android API 16+, but not enabled until API 20+.

jamesdgrant:
Fixes #744

This change patches the security provider, which is recommended [here](https://developer.android.com/training/articles/security-gms-provider) and has the side effect of resolving the connection issue for Android 4.4 and below (issue #744).

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -221,6 +221,8 @@ dependencies {
 
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.facebook.react:react-native:+"  // From node_modules
+    implementation "com.google.android.gms:play-services-basement:$playServicesVersion"
+    implementation "com.google.android.gms:play-services-base:$playServicesVersion"
 
     if (enableHermes) {
         def hermesPath = "../../node_modules/hermes-engine/android/";
--- a/android/app/src/main/java/fr/coopcycle/MainApplication.java
+++ b/android/app/src/main/java/fr/coopcycle/MainApplication.java
@@ -1,12 +1,17 @@
 package fr.coopcycle;
 
-import android.app.Application;
 import android.content.Context;
+import android.content.Intent;
+
 import com.facebook.react.PackageList;
 import com.facebook.react.ReactApplication;
 import com.facebook.react.ReactNativeHost;
 import com.facebook.react.ReactPackage;
 import com.facebook.soloader.SoLoader;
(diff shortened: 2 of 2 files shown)


alexsegura:
Hi @jamesdgrant thank you for this 🙂 

@vladimir-8 can you review it please? I have no skills in Android.

jamesdgrant:
I've tested this change by running the app on Android Studio with an emulated Nexus 4 x86 API 19, target Android 4.4 (with Google APIs). Without the change, there is a "Network request failed" after entering the demo.coopcycle.org server. With the change, the logs show the security provider is updated and the app works normally.

I also tested the app both before the change and after the change using an emulated Pixel 2 x86 API 30, target Android 10.0+ (Google Play), and it works the same each time.

vladimir-8 (approved):

PR #788, opened 2020-09-25
use shared frontend library

vladimir-8:
adds the code from the [shared library](https://github.com/coopcycle/coopcycle-frontend-js) to show colored line near linked tasks:

Screenshot_1599053900

refactoring: some dispatch selectors are moved and used from the [shared library ](https://github.com/coopcycle/coopcycle-frontend-js)


alexsegura:
Thank you @vladimir-8, very nice, refactoring + new feature at the same time 🙂

PR #895, opened 2020-12-11
use v3.0.0 (shared library)

vladimir-8:
breaking changed in the shared library:
- rename lasmile -> logistics
- use '@id' as a key for taskList entities


alexsegura:
Thank you @vladimir-8

PR #1226, opened 2021-09-21
Update Readme

lucasferraro:
I've added some useful links and comments which I've used and followed during my local environment setup.


alexsegura:
Thank you @lucasferraro

PR #1254, opened 2021-12-06
Guest checkout

lucasferraro:
- [x] Add a button to choose if user wants to continue as a guest when is buying.
- [x] If continue as a guest then we ask for an email to send order updates.
- [x] Allow to open order links in app (so guests can check order updates in app when they click in the link that is sended by email)

https://user-images.githubusercontent.com/4819244/144914120-dba6e26d-b282-494e-8391-eec3a53ef5cc.mp4

https://user-images.githubusercontent.com/4819244/144914128-8811aa7a-2ff7-4229-b4bd-79d8b10f4151.mp4

### DIFF
--- a/src/navigation/checkout/Login.js
+++ b/src/navigation/checkout/Login.js
@@ -31,13 +31,17 @@ class Login extends Component {
 
     return (
       <Center flex={ 1 }>
-        <View style={{ paddingHorizontal: 40, width: '100%' }}>
-          <Button colorScheme="success" onPress={() => this.props.guestModeOn()}>
-            { this.props.t('CHECKOUT_AS_GUEST') }
-          </Button>
-        </View>
+        {this.props.guestCheckoutEnabled &&
+          <>
+            <View style={{ paddingHorizontal: 40, width: '100%' }}>
+              <Button colorScheme="success" onPress={() => this.props.guestModeOn()}>
+                { this.props.t('CHECKOUT_AS_GUEST') }
+              </Button>
+            </View>
 
-        <Text mt="4">{ this.props.t('OR') }</Text>
+            <Text mt="4">{ this.props.t('OR') }</Text>
+          </>
+        }
 
         <View style={{ padding: 20 }}>
           <Text style={{ textAlign: 'center' }} note>
@@ -76,6 +80,7 @@ function mapStateToProps(state) {
     message: state.app.lastAuthenticationError,
     isAuthenticated: selectIsAuthenticated(state),
     registrationErrors: state.app.registrationErrors,
+    guestCheckoutEnabled: state.app.settings.guest_checkout_enabled,
   }
 }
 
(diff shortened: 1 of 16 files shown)


alexsegura:
@lucasferraro just a small warning about #1262, that I hotfixed in 927c64d117bfb5c1416090a729751f35944404ec.
I foresee that this can conflict with guest checkout, because as a guest, we don't have any previous orders list to load, nor the user should access to the "my account" screen. I guess the only reliable solution to make the order screen work in both cases, is to [override the back button](https://reactnavigation.org/docs/header-buttons/#overriding-the-back-button)

src/navigation/checkout/Login.js
```diff
@@ -31,6 +31,14 @@ class Login extends Component {
 
     return (
       <Center flex={ 1 }>
+        <View style={{ paddingHorizontal: 40, width: '100%' }}>
+          <Button colorScheme="success" onPress={() => this.props.guestModeOn()}>
+            { this.props.t('CHECKOUT_AS_GUEST') }
```
alexsegura, line 36: Guest checkout can be enabled/disabled per instance. 
The setting should be exposed by the [settings endpoint](https://github.com/coopcycle/coopcycle-web/blob/db50f39d25692ef0d33cb5cc426a870658ba69e1/src/Action/Settings.php), and the app should show this only if enabled.

alexsegura (changes requested):

lucasferraro:
> @lucasferraro just a small warning about #1262, that I hotfixed in [927c64d](https://github.com/coopcycle/coopcycle-app/commit/927c64d117bfb5c1416090a729751f35944404ec). I foresee that this can conflict with guest checkout, because as a guest, we don't have any previous orders list to load, nor the user should access to the "my account" screen. I guess the only reliable solution to make the order screen work in both cases, is to [override the back button](https://reactnavigation.org/docs/header-buttons/#overriding-the-back-button)

@alexsegura nice catch, I'm working in a solution using the `headerLeft` mentioned in the link that you've shared. I'll let you know when it's pushed into this PR. Thanks

src/navigation/checkout/Login.js
lucasferraro: Ok, I'll add it. Thanks

lucasferraro:
> @lucasferraro I tested it, and when you land on the order confirmation screen, after ordering as guest, and hit back, you go through the order list page, which is empty. So, at least, it's not breaking, it's just that this screen may be confusing for customers. It won't be able to load orders list, as the user won't be authenticated, that's all.

Yes, it's not breaking but it's weird to show the orders list for not authenticated/guest users.
I'm working in followingthis approach:

https://user-images.githubusercontent.com/4819244/146037842-71ad486b-5c6f-40b0-9a74-f769cdb35775.mp4

When user is not authenticated the back button redirects to home. But I don't know why after that navigation to home srceen when I try to navigate to "account" section to log in or register for example it goes to order summary :thinking:

alexsegura (approved):
Tested all 3 use cases:
- Checkout as guest ✅ 
- Checkout unauthenticated, and authenticate during checkout ✅ 
- Checkout already authenticated ✅

PR #1279, opened 2021-12-23
Closed restaurants improvements

lucasferraro:
Changes in APP to improve how we show and handle the different posibilites of a closed restaurant.

- **(1)** List of restaurants, when a restaurant is closed we show an icon, the label 'closed now' and the background has less opacity. In this case we have these restaurants:
  - La isla oriental: opened
  - La berenjena: closed but opens today later
  - La silla caliente: closed but opens tomorrow
  - La sarten que fuma: closed and opens another day that is not today or tomorrow
  - La patata caliente: closed with a closing rule, not available
https://user-images.githubusercontent.com/4819244/147282063-eea8a783-7ede-448e-999e-52208417f8ff.mp4

- **(2)** Restaurant page, when it's closed but opens today:

- **(3)** Restaurant page, when it's closed but opens tomorrow and is available "pre order" option:

- **(4)** Restaurant page, when it's closed but opens tomorrow, is not available "pre order" option so the selection on menu is disabled because is not available to make orders:

- **(5)** Restaurant page, when it's closed but opens other day which is not today or tomorrow  and is available "pre order" option:

- **(6)** Restaurant page, when it's closed but opens other day which is not today or tomorrow  and is not available "pre order" option so the selection on menu is disabled because is not available to make orders:

- **(7)** Restaurant page, when it's closed and won't be open until finish a configured closing rule:

### DIFF
--- a/src/navigation/checkout/Restaurant.js
+++ b/src/navigation/checkout/Restaurant.js
@@ -4,6 +4,7 @@ import { connect } from 'react-redux'
 import { withTranslation } from 'react-i18next'
 import { Text } from 'native-base';
 import _ from 'lodash'
+import moment from 'moment'
 
 import CartFooter from './components/CartFooter'
 import AddressModal from './components/AddressModal'
@@ -12,6 +13,7 @@ import ExpiredSessionModal from './components/ExpiredSessionModal'
 import Menu from '../../components/Menu'
 
 import { init, addItem, hideAddressModal, resetRestaurant, setAddress } from '../../redux/Checkout/actions'
+import DangerAlert from '../../components/DangerAlert';
 
 const GroupImageHeader = ({ restaurant, scale }) => {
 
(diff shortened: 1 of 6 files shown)


src/navigation/checkout/Restaurant.js
```diff
@@ -48,18 +55,50 @@ class Restaurant extends Component {
...
+    }
+  }
+
   render() {
 
     const { navigate } = this.props.navigation
     const restaurant = this.props.route.params?.restaurant
     const { isCartEmpty, menu } = this.props
 
+    restaurant.isAvailable = this.state.isAvailable
```
alexsegura, line 92: How can this work on a `const`? 🤔

lucasferraro: Because is an object. You can declare an object as a `const` and is possible to make changes on its properties/fields. What you can't do is re assign that variable.
https://stackoverflow.com/a/59516731

PR #1292, opened 2022-01-04
Added intents to allow map and call apps for Android 11/SDK 30

lucasferraro:
Due to the new [Package Visibilty](https://developer.android.com/training/package-visibility) security feature we need to update the `AndroidManifest.xml` to explicitly allow querying for open map, phone and email apps.

References:
https://github.com/flexible-agency/react-native-map-link/issues/177
https://stackoverflow.com/questions/68659088/intent-action-dial-does-not-function-in-android-11
https://github.com/facebook/react-native/issues/30909#issuecomment-972224894


alexsegura (approved):

alexsegura:
Thank you @lucasferraro 🥳

PR #1311, opened 2022-01-24
Show tabs for multiple instances in the same city

### LINKED ISSUE #1295: Improve management of multiple instances in the same city
Currently, as we display city names to the customer, when there are multiple co-ops in the same city, we display "City - Name of co-op", which can be confusing when there is a lot (like in Mexico City). 

Instead, at least we should display the city name, and open a popup to choose the co-op. 

A better implementation would be to allow switching between co-ops in the same city. 
It could be implemented by parsing the `coopcycle.json` data, grouping co-ops by the same city name, and then displaying "tabs" on the restaurant list screen. Clicking on each tab would load the restaurant list of each co-op, and customers would be able to navigate this way. At some point, when clicking on a restaurant for example, the app should connect to the instance under the hood, to keep the current behavior. 

---

Tabs

alexsegura:
Fixes #1295

### DIFF
--- a/src/Server.js
+++ b/src/Server.js
@@ -47,6 +47,20 @@ const save = data => {
   })
 }
 
+function overrideCity(values) {
+
+  return values.map((value) => {
+    if (value.city.startsWith('Ciudad de México')) {
+      return {
+        ...value,
+        city: 'Ciudad de México'
+      }
+    }
+
+    return value
+  })
+}
+
 class Server {
 
   static loadAll() {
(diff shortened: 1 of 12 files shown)


alexsegura:
@lucasferraro here is the branch I started. The commit 3b25e203e21128a29b506d505afae63ab2ee4d1d should be removed before merging.

lucasferraro:
> @lucasferraro here is the branch I started. The commit [3b25e20](https://github.com/coopcycle/coopcycle-app/commit/3b25e203e21128a29b506d505afae63ab2ee4d1d) should be removed before merging.

We've decided don't remove that commit to ensure backward compatibility with these changes.

alexsegura:
Thank you @lucasferraro, it looks perfect with "real" tabs 🙂

PR #1340, opened 2022-02-14
New packages and weight section in task screen

lucasferraro:
With these changes now we are showing in the pickup task all the packages and the weight from all subsequent dropoffs.
Also, on each dropoff task now we show the packages and weight to be delivered.

https://user-images.githubusercontent.com/4819244/153931540-57badfeb-cffb-42e2-a35f-7060ad3f86c0.mp4

For now we keep the packages and weight info in the pickup task comments because if some admin/courier has an old version of the app can still see this info in comments. But after a period, we are going to remove this behaviour and then comments won't include this info.
This is how it looks now with the info in comments too:
</img>


alexsegura (approved):
As a second version, we could use a list of checkboxes (that does nothing), so that messengers can use it as a "checklist". But it's fine for now 🙂

PR #1422, opened 2022-07-22
Naofood ios without update

tetoL44:


ios/Naofood.xcodeproj/project.pbxproj
```diff
@@ -703,14 +702,14 @@
...
-			baseConfigurationReference = 4950D879877C4AFABC75A2ED /* Pods-Naofood.debug.xcconfig */;
+			baseConfigurationReference = 9B9A56EB3BCE389586CBC1FC /* Pods-Naofood.debug.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Naofood/Naofood.entitlements;
 				CURRENT_PROJECT_VERSION = 1;
 				DEFINES_MODULE = YES;
-				DEVELOPMENT_TEAM = 66HUGQ2V5L;
+				DEVELOPMENT_TEAM = LKV8P3U392;
```
alexsegura, line 712: Not sure this should be committed

src/navigation/checkout/SearchForm.js
```diff
@@ -46,6 +46,7 @@ const Autocomplete = (props) => {
 
   return <FlatList
     data={props.restaurants}
+    alwaysBounceVertical={false}
```
alexsegura, line 49: Should this change only apply for iOS? 
If yes, you can use conditionals `alwaysBounceVertical={ Platform.OS === 'ios' ? false : true }`

src/navigation/checkout/Summary.js
```diff
@@ -164,6 +165,7 @@ class Summary extends Component {
     return (
       <FlatList
         data={ this.props.cart.items }
+        alwaysBounceVertical={false}
```
alexsegura, line 168: Should this change only apply for iOS?
If yes, you can use conditionals `alwaysBounceVertical={ Platform.OS === 'ios' ? false : true }`

alexsegura (commented):
I see that the `yarn.lock` has many updates but nothing has changed in `package.json`. 
I guess you ran `yarn upgrade`? Does it still work on Android?

alexsegura:
To manage multiple `GoogleService-Info.plit` files --> https://bamtech.gitbook.io/dev-standards/react-native/setup/setup_firebase_multiple_envs.mo

PR #1423, opened 2022-07-27
Support multiple carts and multiple instances

lucasferraro:


vladimir-8:
Currently not planned; closing it for now.

PR #1483, opened 2023-03-13
Bulk operations on tasks

### LINKED ISSUE #1476: As a flying dispatcher/courier, I want to select multiple tasks in an easier way to save time
Right now, as a courier or a dispatcher, I can only select one task at a time before choosing what to do with it.

With the multiplication of routes with large numbers of points, I would need to be able to handle quicker this important number of tasks.

1- As a courier, having the possibility to select multiple pick ups at same location and finishing these tasks in one go would save a lot of time.

2-As a flying dispatcher, I could select multiple tasks and assign them to a courier directly without having to do it one by one.

I found this model on another solution. Maybe checkboxes is another option

image

lucasferraro:
Closes #1476 

**Assign multiple tasks at once**
_As a flying dispatcher, I could select multiple tasks and assign them to a courier directly without having to do it one by one._
https://user-images.githubusercontent.com/4819244/224722520-df4cb854-2da3-48d8-9a27-311e4b2798a6.mp4

**Complete multiple tasks at once**
_As a courier, having the possibility to select multiple pick ups at same location and finishing these tasks in one go would save a lot of time._
https://user-images.githubusercontent.com/4819244/224722666-039e570a-183a-4f63-bb8c-50c0684e89b6.mp4

**Bonus track**
Allow to attach images/signatures to multiple tasks and complete
https://user-images.githubusercontent.com/4819244/224722887-e8d31f49-3a64-48df-b5de-f7e66fcd0564.mp4


alexsegura (approved):

ERAMANrepartos:
So apparently this feature is having issues with the implementation because the long press does not work on old android phones, and also does not work on iOS phones. I talked with @alexsegura and we have both seen that users, when asked to try to validate multiple tasks, as seen in the video below

https://github.com/coopcycle/coopcycle-app/assets/77277854/f0df28f2-281f-4594-99d2-71e787bcd9dc

We propose switching to using this as the way to select multiple tasks instead of long press. 
what do you think @lucasferraro?

ERAMANrepartos:
Two things to resolve:

1. Do we want to allow a courier to close a pickup and a dropoff at the same time? In a way this sort of encourages a courier in the bad habit of not closing tasks in the appropriate moment. I personally think we shouldn´t limit the functionality, especially since this seems like a general responsibility of the cooperative to enforce, and there are a million ways to no correctly close the tasks anyways
2. What do we want to happen if there is an error? For example, if you try to close 5 dropoffs, but one of them is associated with a still unfinished pickup? Ideally it would 
     a. Successfully close all the tasks that can be closed
     b. Show an error that "task #XXXXX associated with order YYYY could not be closed because the related pickup was not completed yet"
     c. Leave this tasks unfinished. 
     
How does that seem @lucasferraro ?

lucasferraro:
Here my answers:

1. I agree with is better to don't limit the functionality and allow to close a pickup and dropoff at the same time.
2. I think that the flow proposed could be nice for the user experience. I'm attaching a video so you can see how I'm implementing it and if it's ok for you this could be ready to push to tech review @ERAMANrepartos @alexsegura :


https://github.com/coopcycle/coopcycle-app/assets/4819244/2f8bcf4b-9f13-43a3-88d6-5122fbcf3967

Notes:
- Task # 900 is dropoff and the pickup related to this (# 899) is already completed -> This task is successfully closed
- Tasks # 896 and # 897 both are dropoff and the pickup related to them (# 895) is not completed -> These tasks couldn't be closed and a message is shown explaining why.

Issue #1472, opened 2023-01-24
I want the application add an image to a task faster, or add it in such a way that it does not impede my work as a courier


rayon9coop:
**Describe the bug**
All is in the title :-)
except for the fact that this problem appeared with the release of version 2.5.1.

**To Reproduce**
1. In parameters, deselect "Default signature screen". 
2. Open a task
3. Click on "finish" and then on "Add a POD"
4. count the seconds (8 with my smartphone)
5. enable the flash and take the pic
6. count the seconds (9 with my smartphone)

**Screenshots**

https://user-images.githubusercontent.com/44605344/214253679-e37bef70-03c7-4880-886c-d1cd327928cd.mp4

Smartphone (please complete the following information):

Device: [Samsung XCOVER Pro] SM-G715FN/DS
OS: [Android 13 - ONE UI 5.0]
app 2.5.1

ERAMANrepartos:
I understand that the solution is zipping photos before the are uploaded given past conversations. @alexsegura ?

alexsegura:
We are using Base64 encoding to manipulate files. It means the image is converted into a huge string and passed around, before being re-converted into a byte stream. This can lead to [poor performance](https://github.com/RonRadtke/react-native-blob-util#performance-tips). 
It seems possible to use the filesystem/image URIs instead.

ERAMANrepartos:
@alexsegura Does using filesystem/image URIs make compressing unnecessary? Sometimes we do upload a dozen photos at a time

ERAMANrepartos:
This can also be accomplished by offline uploading of tasks. Check #1492

alexsegura:
> Does using filesystem/image URIs make compressing unnecessary?

@ERAMANrepartos there are multiple vectors for optimization. Using the filesystem should improve responsiveness of the app, but it can't reduce the size of data that needs to be transferred. Actually, we are already telling the camera to [take photos at 50% quality](https://github.com/coopcycle/coopcycle-app/blob/3db496aaa0c1d32aafa11ba17a9e2e82922b243d/src/navigation/task/Photo.js#L68), so the images should already be not so big.

ERAMANrepartos:
What i would like to do it make individual issues for each vector we want to actually implement. Ill make an issue for 50% resolution photos. Any other issue we should create @alexsegura ?

PR #1493, opened 2023-05-09
Improve photos performance.

alexsegura:
See #1472.

- Use react-native-blob-util instead of rn-fetch-blob.
- Avoid using Base64.


alexsegura:
I still have a problem when adding a signature

Capture d’écran 2023-05-10 à 16 37 09

alexsegura:
@r0xsh can you also try with signatures?

r0xsh (approved):

Issue #1455, opened 2022-11-08
Linked recurring tasks: the colours of the sidebar seem random instead of being the same. 


rayon9coop:
**Describe the bug**
On the courriers app, the colors of linked reccuring task aren't the same as they are on the dispatch pannel.

**To Reproduce**

1. Create reccuring tasks with at least 3 tasks
2. Generate tasks
3. Assign these tasks
4. See on dispatch pannel the color of side bar of linked tasks
5. See in the app the color of these tasks seems random

**Expected behavior**
The same colors in the dispacth screen and in the courriers app

**Screenshots**
On admin pannel:
![image](https://user-images.githubusercontent.com/44605344/200548242-24698281-96a9-4ee4-bfe1-71ebaaa306ef.png)

In the app
![Screenshot_20221108-115407_CoopCycle](https://user-images.githubusercontent.com/44605344/200583971-9d8173d6-6528-47c6-a4cc-01d29ffd6c46.jpg)
![Screenshot_20221108-115414_CoopCycle](https://user-images.githubusercontent.com/44605344/200584105-e19be7aa-fc7f-4d3d-a53e-04ad2b1443dc.jpg)
![Screenshot_20221108-115428_CoopCycle](https://user-images.githubusercontent.com/44605344/200584257-f192a022-04e3-43dd-bf6a-e3a2b66300b3.jpg)
![Screenshot_20221108-115438_CoopCycle](https://user-images.githubusercontent.com/44605344/200584302-d7a4868e-a04d-4ba8-8a6e-b0a1547698b5.jpg)

vladimir-8:
This happens because the app uses the old version of the https://github.com/coopcycle/coopcycle-frontend-js library. The latest version needs to be integrated into the repository using https://github.com/ingydotnet/git-subrepo (the same way it's done on the https://github.com/coopcycle/coopcycle-web). I'll try to fix it

PR #1507, opened 2023-05-17
Fix: linked tasks colour

vladimir-8:
- Integrated the latest version of the https://github.com/coopcycle/coopcycle-frontend-js library (using https://github.com/ingydotnet/git-subrepo)
- Fixed: https://github.com/coopcycle/coopcycle-app/issues/1455


r0xsh (commented):
Thank you for contribution and proposing changes. However, there is a minor adjustment needed in the pull request. Instead of using git submodule, we need to update the version of the `coopcycle-frontend-js` in package.json. Apart from that, everything else looks good. Excellent job, and thank you once again!

See: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#git-urls-as-dependencies

vladimir-8:
Thanks for the review! 

That's on purpose, actually. I remember discussing it with @alexsegura at some point. We concluded that it is way easier to develop a library when the code is in the same folder rather than releasing a new version and importing it as a dependency in package.json. The same approach is used in [github.com/coopcycle/coopcycle-web/](https://github.com/coopcycle/coopcycle-web/) If sticking to this approach, it could be removed from package.json dependencies.

vladimir-8:
Great! Thanks for reviewing @lucasferraro! Will you merge it, or should I?

lucasferraro:
> Great! Thanks for reviewing @lucasferraro! Will you merge it, or should I?

Merged ;)

PR #1527, opened 2023-07-23
Delivery form

### LINKED ISSUE #1522: Delivery form in app
As mentioned in #3672  I would like to request the implementation of an in-app delivery request feature to enable users to place delivery deliveries ("encargos") from point A to point B within the CoopCycle app. This functionality is available through a delivery form or as a user with a store account, but it is not directly accessible from the app itself.

**In order to**  `achieve something`
The objective is to provide a more streamlined and convenient way for users to request orders through the CoopCycle app.

**As a** `role`
As a final user of the CoopCycle app.

**I want** `something to be added or changed`
I want to have a way to set up deliveries directly from the app. Currently, this can be done through a form on the homepage, but it should also be accessible within the app.

**Whereas currently** `this happens`
Currently, a form setup on the homepage allows for delivery requests, but it is not integrated into the app itself. This form has been partially implemented by @AwesomeKuro 

**Additional context**

- The form can be included through the left lateral menu
- It should be possible to set up an specific pricing for the delivery requests
- The flow of item deliveries should imitate that of popular ride-hailing services

Additional context:
This feature request has been inspired by feedback and input from various courier services in Mexico. 
@AwesomeKuro  can implement this with @alexsegura and @@Paul-Eraman-CoopCycle .

Thank you  😀.

AwesomeKuro:
Receives the delivery form URL from settings API and sets it as the link opened when entering the newly added Delivery section on the sidebar menu.

Fixes #1522

### DIFF
--- a/src/navigation/delivery/Delivery.js
+++ b/src/navigation/delivery/Delivery.js
@@ -16,12 +16,6 @@ class Delivery extends Component {
 }
 
 function mapStateToProps(state) {
-  const publicServers =
-    _.filter(state.app.servers, s => !!s.coopcycle_url)
-
-  const showAbout =
-    _.includes(publicServers.map(s => s.coopcycle_url), state.app.baseURL)
-
   return {
     defaultDeliveryFormUrl: state.app.settings.default_delivery_form_url,
   }
--- a/src/navigation/delivery/Delivery.js
+++ b/src/navigation/delivery/Delivery.js
@@ -7,6 +7,7 @@ import _ from 'lodash';
 
 class Delivery extends Component {
   render() {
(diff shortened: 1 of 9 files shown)


src/navigation/delivery/Delivery.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+}
+
+function mapStateToProps(state) {
+  const publicServers =
+    _.filter(state.app.servers, s => !!s.coopcycle_url)
+
+  const showAbout =
+    _.includes(publicServers.map(s => s.coopcycle_url), state.app.baseURL)
+
```
alexsegura, line 24: Dead code

PR #1533, opened 2023-08-16
Removed override for CDMX

AwesomeKuro:
Navigation was not very useful in the restaurants view, and TWC being inactive broke navigation into CDMX servers.


alexsegura:
It is not what was expected, we still want all restaurants to appear under same city. Closing.

PR #1544, opened 2023-09-08
Show payment method info on tasks and orders

### LINKED ISSUE #1542: To [see the method of payment for each order] As a [restaurant or courier] I want to [show the payment method icons]
**For** `see the method of payment for each order`
In Mexico we accept cash payment, so the courier and the restaurant need to know what the payment method was for each delivery.

**As** `restaurant or courier`
the courier and the restaurant need to know this information without having to use the computer to properly handle the cash

**I want** `show the payment method icons`
for the courier, on the task list screen, show in each order what is his payment method and for the restaurant, on the main screen or in the order details.

**While currently** "this happens"
the restaurant has to open the web version or ask the dispatcher and the courier always has to ask the dispatcher

**Additional context**
Since we are handling cash, when the courier charges the customer, sometimes they do not have the money to settle the payment, that is, the order was $430, the customer pays with a bill of $500, if the courier does not have the $70 , is a problem, another improvement would be that the client had a field to tell us how much money he will pay with
Imagen de WhatsApp 2023-09-07 a las 08 42 25
Imagen de WhatsApp 2023-09-07 a las 08 42 251
Imagen de WhatsApp 2023-09-07 a las 08 42 252
issue1
issues

lucasferraro:
Closes #1542 

**For courier**
- task list screen -> PM icon

- task detail screen -> PM icon + description

 
**For restaurant**
- order list -> PM icon

- order detail -> PM icon + description

### DIFF
--- a/src/components/TaskListItem.js
+++ b/src/components/TaskListItem.js
@@ -15,6 +15,7 @@ import {
   taskTypeIconName,
 } from '../navigation/task/styles/common'
 import TaskTitle from './TaskTitle'
+import { PaymentMethodInfo } from './PaymentMethodInfo'
 
 const styles = StyleSheet.create({
   itemIcon: {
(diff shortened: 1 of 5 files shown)


src/components/TaskListItem.js
```diff
@@ -213,7 +214,13 @@ class TaskListItem extends Component {
               <Text numberOfLines={ 1 } style={ textStyle }>{ task.address.streetAddress }</Text>
               <HStack alignItems="center">
                 <Text pr="2" style={ textStyle }>{ moment(task.doneAfter).format('LT') } - { moment(task.doneBefore).format('LT') }</Text>
-                { task.comments && task.comments.length ? <Icon as={FontAwesome} name="comments" size="xs"/> : null }
+                { task.comments && task.comments.length ? <Icon mr="2" as={FontAwesome} name="comments" size="xs"/> : null }
+                {
+                  task.metadata && task.metadata.payment_method &&
```
alexsegura, line 219: `task.metadata.payment_method` may be an empty string, so it would be better to use a ternary operator to avoid [this bug ](https://github.com/facebook/react-native/issues/29980)

alexsegura: Ok actually the bug doesn't happen even with an empty string, no problem 🙂

PR #1547, opened 2023-09-19
Add process.env condition for sentry init

Pierre-LHOSTE:


src/App.js
```diff
@@ -21,9 +21,11 @@ import Smartlook from 'react-native-smartlook-analytics'
 import Config from 'react-native-config'
 import * as Sentry from '@sentry/react-native'
 
-Sentry.init({
-  dsn: Config.SENTRY_DSN,
-})
+if (process.env.NODE_ENV === 'production') {
```
alexsegura, line 24: I would use `if (!__DEV__)` as in other parts of the code.

r0xsh: I was not aware that a constant were existing.
He did the condition this way based on a feedback i gave him ;
> We should flip the condition the other way because developers often use multiple environment dev names such as `dev`, `devel`, `development`, `test`, `staging` etc.
And since we only want sentry enable on production, you should check for the production environment name

r0xsh: Okay by looking at the React Native code i found this
https://github.com/facebook/react-native/blob/f39952ba78d4b7d867c01eabffdf6252919ff52f/packages/react-native/Libraries/Core/setUpGlobals.js#L32-L35

So yeah i guess it's safe to say that we can use the `if (!__DEV__)` sorry @Pierre-LHOSTE for my misleading feedback (:

r0xsh, line 24:
```suggestion
if (!__DEV__) {
```

r0xsh (changes requested):

PR #1549, opened 2023-09-23
feat: `OrderTracking` now parse new `events` key on orders

r0xsh:
Depends on: coopcycle/coopcycle-web#3774

### DIFF
--- a/src/navigation/checkout/OrderTracking.js
+++ b/src/navigation/checkout/OrderTracking.js
@@ -14,26 +14,39 @@ import Step from './components/Step';
 import { find } from 'lodash'
 
 
-function stateToStep(state) {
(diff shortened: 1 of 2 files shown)


src/navigation/checkout/OrderTracking.js
```diff
@@ -101,7 +106,7 @@ class OrderTrackingPage extends Component {
       moment(order.shippingTimeRange[1]).format('HH:mm'),
   ]
 
-    const step = stateToStep(order.state)
+    const step = stateToStep(order.events)
```
alexsegura, line 109: The problem here is that if the server does not return this property, it will break. It should be backwards compatible.

Issue #1546, opened 2023-09-13
When completing a task as failed, allow selecting between a list of "reasons"


alexsegura:
When the messenger completes a task as failed, it should show a pre-defined list of explicit reasons for the failure. 
For example, "customer is not at home", "customer didn't answer to the phone", etc...

When showing the task failure screen, retrieve the list of reasons associated to this task (see coopcycle/coopcycle-web#3768), and send it with the payload to the API.

PR #1566, opened 2023-10-04
Add failure reason input on failed task screen

r0xsh:
image

Fixes: https://github.com/coopcycle/coopcycle-app/issues/1546

### DIFF
--- a/src/navigation/task/Complete.js
+++ b/src/navigation/task/Complete.js
@@ -78,11 +78,11 @@ const AttachmentItem = ({ base64, onPressDelete }) => {
 
 const FailureReasonPicker = ({ task, httpClient, onValueChange }) => {
 
-  const { data, isSuccess } = useQuery([ 'task', 'failure_reasons', task['@id'] ], async () => {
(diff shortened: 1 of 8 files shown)


r0xsh:
Depends on: coopcycle/coopcycle-web#3792

r0xsh:
Fixes: #1546

src/navigation/task/Complete.js
```diff
@@ -53,19 +76,54 @@ const AttachmentItem = ({ base64, onPressDelete }) => {
   )
 }
 
+const FailureReasonPicker = ({ task, httpClient, onValueChange }) => {
+
+  const { data, isSuccess } = useQuery([ 'task', 'failure_reasons', task['@id'] ], async () => {
+    return await httpClient.get(`${task['@id']}/failure_reasons`)
```
alexsegura, line 82: To be compatible with older servers, when this endpoint returns a 404, the reason picker shouldn't appear.

r0xsh: Tried to hide it if 404 is returns but it was looking clunky, so i preferred to add an error message instead
image

PR #1579, opened 2023-10-18
Lista mensajerias

AwesomeKuro:
Now, override city has been modified to check the server almost as if i clicked it (it doesn't save settings, but still is slow) and prevents the unavailable servers from showing up on the selection list. CDMX no longer has a submenu with multiple servers, probably because only TIG is working


alexsegura (changes requested):
Can you please remove `console.log` statements?

AwesomeKuro:
> Can you please remove `console.log` statements?

i removed them all except one handling the error, should that one also be deleted?

PR #1661, opened 2023-12-28
Fix search crash

r0xsh:


vladimir-8:
While this PR fixes the crash, I think we should create another issue to actually allow a customer to open all found restaurants and products from the search results, and not only the ones that we have in the redux

vladimir-8 (approved):
While this PR fixes the crash, I think we should create another issue to actually allow a customer to open all found restaurants and products from the search results, and not only the ones that we have in the redux

alexsegura:
I removed the "Fixes"

r0xsh:
> While this PR fixes the crash, I think we should create another issue to actually allow a customer to open all found restaurants and products from the search results, and not only the ones that we have in the redux

Initially, I solved the problem by lazy loading the restaurant information. But then, I noticed it's strange to display a restaurant where you can't place an order. If this is what we want, we need to change the Restaurant page to prevent order creation. Because, as I experienced yesterday, when you try to place an order, the app immediately tells you that delivery isn't possible, which is odd.

r0xsh:
Or maybe show all the restaurants in search results but the one that can't be ordered from are disabled (like putting them in a grey color)

Issue #1670, opened 2024-01-03
In order to make code easier to understand I want remove semi-arbitrary delay while hiding the loading indicator


vladimir-8:
**I want** `something to be added or changed`
A clear and concise description of what is to be added or changed in the software

Based on the tests I did, it seems that after removing the delay for a spinner, everything still works as expected. I hope that this delay is not needed anymore. However, I’m not entirely confident it will not put back this issue: https://github.com/coopcycle/coopcycle-app/issues/1265. 

I propose to make this change in a few steps:

- [ ] 1. Ask a few couriers to manually set it via the feature flags

- [ ] 2. Set it as a default option, while still allowing to revert it via a feature flag
https://github.com/coopcycle/coopcycle-app/blob/a67eee6cc3cd6a6b51968dc6340744e177a2f213/src/redux/App/reducers.js#L100
https://github.com/coopcycle/coopcycle-app/blob/a67eee6cc3cd6a6b51968dc6340744e177a2f213/src/redux/App/selectors.js#L145

- [ ] 3. Remove the feature flag and clean up the code

**Whereas currently** `this happens`
A clear and concise description of what happens now

I suspect that the issue with displaying Modals on iOS (for example, maintenance message: https://github.com/coopcycle/coopcycle-app/issues/1669) is happening due to this extra delay that we added before hiding a spinner: https://github.com/coopcycle/coopcycle-app/commit/866fd9a8ce4c65c52ccb5d8270d380df2882d87a

https://coopcycle.slack.com/archives/G01JR0YF5V3/p1704250963518979

vladimir-8:
Currently, the delay is applied only in the Courier interface on iOS: https://github.com/coopcycle/coopcycle-app/blob/master/src/redux/App/selectors.js#L176-L182 Removing it causes the app to be stuck (invisible Modal overlay) on task completion. We need to figure out why that happens before we can remove it. That has low priority, however.

Issue #1669, opened 2023-12-29
Maintenance message does not appear on iOS


vladimir-8:
**To Reproduce**
1. Enable maintenance mode on any of the instances with custom apps or an other instance that you control
2. (optional) create an app build with a hard-coded DEFAULT_SERVER env variable pointing to an instance from 1
3. Open the app
4. Enter the address
5. See there is no maintenance message

**Expected behavior**
The message should be shown, see the Android app:

![Screenshot_1703877101](https://github.com/coopcycle/coopcycle-app/assets/70273239/8ce39d22-014e-4fa2-bd72-5823bf73815f)

**Screenshots**
![Simulator Screenshot - iPhone SE (3rd generation) - 2023-12-29 at 11 07 55](https://github.com/coopcycle/coopcycle-app/assets/70273239/928dd450-e1ff-4f71-b632-8a9770eb432a)

**Smartphone (please complete the following information):**
 - OS: iOS

**Additional context**

There could be a problem with displaying modals in general. Here is where we show them: https://github.com/coopcycle/coopcycle-app/blob/c1cdb5f1ed0c5eb02dd5acd05d1712a5030484cf/src/navigation/Loading.js#L88

Issue #1265, opened 2021-12-13
Never ending loader on task list screen


alexsegura:
When looking at server logs, we see that requests to `/api/me/tasks/{date}` are sent twice, i.e the server always receives 2 requests to this endpoint at the same minute/second, and sometimes it returns an HTTP 500 error, because of a race condition. 

This could explain the bug where messengers see a never ending loader on the tasks screens. 
Maybe we should implement a queue when calling the Redux `loadTasks` action?

<img width="1106" alt="Capture d’écran 2021-12-13 à 11 37 29" src="https://user-images.githubusercontent.com/1162230/145797420-bfbae9b1-8ad0-4689-a6b5-b5d11ebc3453.png">

alexsegura:
People are reporting its impacting only iPhones. 
We need to check on server logs to see if the double requests are coming from iPhone user agents.

PR #1671, opened 2024-01-03
Maintenance message and loading indicator updates

vladimir-8:
1. A hotfix to show maintenance message on iOS: https://github.com/coopcycle/coopcycle-app/issues/1669
2. An experimental implementation without a delay, behind a feature flag: https://github.com/coopcycle/coopcycle-app/issues/1670

### DIFF
--- a/src/components/Spinner.js
+++ b/src/components/Spinner.js
@@ -2,7 +2,10 @@ import React, { Component, PureComponent } from 'react'
 import { connect } from 'react-redux'
 import Spinner from 'react-native-loading-spinner-overlay'
 
-import { selectIsLoading } from '../redux/App/selectors'
+import {
+  selectIsLoading,
+  selectIsSpinnerDelayEnabled,
+} from '../redux/App/selectors'
 
 class SpinnerWrapper extends PureComponent {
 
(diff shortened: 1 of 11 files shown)


Atala (commented):
hello, I am not 100% sure of what is expected here + I cant test on iOS so I will trust you on this but

- I don't get why we have to make this configurable with the "tap menu"? you want to test it yourself in production.
- it shows the modal alright on my phone, but only if I quit the app after having set maintenance ON. If I do :

vladimir-8:
>     I don't get why we have to make this configurable with the "tap menu"? you want to test it yourself in production.

Yes, I want to ask a few couriers to test it on production. 

I suspect that maintenance message is not appearing due to this extra delay that we added before hiding a spinner: https://github.com/coopcycle/coopcycle-app/commit/866fd9a8ce4c65c52ccb5d8270d380df2882d87a I really want to avoid fixing it by adding another delay, which just increases the technical debt. Also, based on the tests I did, it seems that after removing the delay for a spinner, everything still works as expected. I hope that this delay is not needed anymore. However, I’m not entirely confident it will not put back this issue: https://github.com/coopcycle/coopcycle-app/issues/1265. What I think would be great to do is to make this change behind a feature flag, make it possible to enable/disable it from the app and then beta-test it with a few couriers before enabling it globally. https://coopcycle.slack.com/archives/G01JR0YF5V3/p1704250963518979


>     it shows the modal alright on my phone, but only if I quit the app after having set maintenance ON. If I do :
>         in the app go to restaurants page
>         on the platform turn maintenance ON
>         the app crashes "Unexpected error occured" and I see in the logs "503 was returned by the server"
>         if I stop the app and launch it again, then the maintenance message is alright
>         -> is that the expected behavior?

Well, it's another issue :)

>     also one cant close the maintenance modal without stopping the app, but I guess it is the expected behavior

And another one

Issue #1639, opened 2023-12-11
The CoopCycle logo appears instead of the kooglof! logo


Paul-Eraman-CoopCycle:
https://github.com/coopcycle/coopcycle-app/assets/77277854/9db32128-8b2e-4a92-b4fe-eb69425710c8

Should be the kooglof logo!

r0xsh:
https://docs.expo.dev/versions/latest/sdk/splash-screen/

Perpolita:
only on iPhones, it's working on Android's phones

vladimir-8:
Will be fixed in https://github.com/coopcycle/coopcycle-app/pull/1672. Updated iOS launch screen to use a solid background color as per Apple guideline https://developer.apple.com/design/human-interface-guidelines/launching#Launch-screens

If we really want to keep the splash screen it should be implemented using a different approach, for example Expo, as suggested by Antoine https://github.com/coopcycle/coopcycle-app/issues/1639#issuecomment-1850463936

PR #1672, opened 2024-01-04
Update iOS launch screen

vladimir-8:


alexsegura:
@vladimir-8 where can I find the Apple guidelines you are refering to? It looks like a small regression to me, we don't have a launchscreen anymore, just a blank screen.

vladimir-8:
@alexsegura https://github.com/coopcycle/coopcycle-app/issues/1639#issuecomment-1877909482 In my opinion, such launch screens are a bit excessive. But I see many apps doing this, so maybe they are onto something. In that case, we can re-implement it using https://docs.expo.dev/versions/latest/sdk/splash-screen/ And make it a bit smarter by hiding it when loading actually finishes

Issue #1600, opened 2023-11-13
In the app a rider can show a simple polyline between their tasks


Atala:
<s>**In order to**  `achieve something`
- When a rider is assigned a tour, I should see the tour "line" on the map view of the tasks list. Or other vizualisation that the tasks are linked
- In the list view of the tasks list nothing is changed</s>

-> changed to "In the app a rider can show a simple polyline between their tasks"

Paul-Eraman-CoopCycle:
This should not be when a rider is assigned a tour, it should be a setting in the configurations like the activation or deactivation of seeing finished tasks

PR #1678, opened 2024-01-05
feat: add ability to show polyline on map

Atala:
issue #1600

- polyline updates when changin task order
- ON by default (so user see at least the feature once)
- save when turning ON/OFF the app

in action [https://coopcycle.slack.com/archives/G01JR0YF5V3/p1704366584889179](https://coopcycle.slack.com/archives/G01JR0YF5V3/p1704366584889179)

### DIFF
--- a/src/components/TasksMapView.js
+++ b/src/components/TasksMapView.js
@@ -7,18 +7,22 @@ import {
   TouchableOpacity,
   View,
 } from 'react-native'
-import { Callout, Marker } from 'react-native-maps'
+import { Callout, Marker, Polyline } from 'react-native-maps'
 import ClusteredMapView from 'react-native-maps-super-cluster'
 import Modal from 'react-native-modal'
 import { withTranslation } from 'react-i18next'
+import { connect } from 'react-redux'
+
 
 import {
   greyColor,
   whiteColor,
+  blueColor
 } from '../styles/common'
 import { uniq } from 'lodash'
 import TaskMarker from './TaskMarker'
 import TaskCallout from './TaskCallout'
+import {selectIsPolylineOn} from '../redux/Courier'
 
 const clusterContainerSize = 40
 
@@ -264,6 +268,14 @@ class TasksMapView extends Component {
             onClusterPress={ this.onClusterPress }
             { ...otherProps }>
             { this.props.children }
+            {
+              this.props.isPolylineOn ?
+              (<Polyline
+                coordinates={this.props.tasks.map(task => ( task.address.geo ))}
+                strokeWidth={3}
+                strokeColor={blueColor}
+             />) : null
+            }
           </ClusteredMapView>
         ) : null }
         { this.renderModal() }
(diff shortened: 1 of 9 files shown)


src/components/TasksMapView.js
```diff
@@ -264,6 +268,14 @@ class TasksMapView extends Component {
...
             { ...otherProps }>
             { this.props.children }
+            {
+              this.props.isPolylineOn &&
+              (<Polyline 
+                coordinates={this.props.tasks.map(task => ( task.address.geo ))}
+                strokeWidth={3}
+                strokeColor={blueColor}
+            ></Polyline>)
+            }
```
vladimir-8, line 278: I suggest using a ternary operator as it's less error prone: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

```
this.props.isPolylineOn ?
(<Polyline 
  coordinates={this.props.tasks.map(task => ( task.address.geo ))}
  strokeWidth={3}
  strokeColor={blueColor}
/>) : null
```

src/redux/Courier/taskSelectors.js
```diff
@@ -17,6 +17,7 @@ export const selectIsTasksRefreshing = state => state.entities.tasks.isRefreshin
 export const selectIsTasksLoadingFailure = state => state.entities.tasks.loadTasksFetchError
 export const selectIsTaskCompleteFailure = state => state.entities.tasks.completeTaskFetchError
 export const selectTaskFilters = state => state.ui.tasks.excludeFilters
+export const selectIsPolylineOn = state => state.ui.tasks.isPolylineOn
```
vladimir-8, line 20: I guess `isPolylineOn` will be `undefined` for the existing users. Is it what we want? Or do we  want it also to be true by default in that case?

vladimir-8 (approved):

src/redux/Courier/taskSelectors.js
Atala: I think the initial state is set here : https://github.com/coopcycle/coopcycle-app/pull/1678/commits/a577f44278b49bb2afe65a69db844bc8b9eb4410#diff-9f556b044c2b86f230999b3167c6717644249d59e90412788a54f7ded0c24c6aR26 so it will be set to true (I have set to true by default so existing users see the feature)

am I correct?

src/components/TasksMapView.js
Atala: OK I will edit

src/redux/Courier/taskSelectors.js
vladimir-8: Maybe it's correct, but I suggest to double check that it works that way for the existing users, and that it's not overridden by the react-persist during rehydration: https://github.com/rt2zz/redux-persist#state-reconciler

Atala: ah ok I see. if it is not correct I have attempted a commit (see below).

however I am not 100% sure on how to test locally? I should build the app on master then switch branch and load this one? through Metro?

thx

PR #1684, opened 2024-01-10
Fix: registration during the checkout flow

vladimir-8:


src/API.js
```diff
@@ -159,29 +160,46 @@ Client.prototype.createRequest = function(method, url, data, options = {}) {
...
+      } else {
+        console.warn(`⬅ ${method} ${uri} | ${error.message}`)
+      }
+
+      return Promise.reject(error)
+    })
 }
 
-Client.prototype.get = function(uri, data, options = {}) {
+Client.prototype.get = function(uri, options = {}) {
```
alexsegura, line 185: What if we want to pass GET query parameters?

vladimir-8: We will need to find a workaround for it. I remember while debugging another issue I noticed that either on Android or iOS GET requests are rejected if they include a body. And we already ignore this parameter for GET requests and only add it to POST and PUT requests: https://github.com/coopcycle/coopcycle-app/blob/master/src/API.js#L146-L159 That's why I thought it will be cleaner to remove it for now, rather than spreading `{}` in the method calls across the whole project

alexsegura: It would be cool to keep the same method signature IMHO. For GET query parameters, Axios accepts the [`params` key in the options object](https://axios-http.com/docs/req_config). In `createRequest`, we could do this:

```javascript
if (method.toUpperCase() === 'GET') {
  req.params = params
}
```

vladimir-8: In my opinion, it's very confusing if for GET requests we start mapping the data field to params. I'd be happy to keep the same method signature, but if `data` field is not applicable to GET requests, I think, it's better to make it explicit, rather than give an impression that it works the same as for other requests.

alexsegura (approved):

PR #1716, opened 2024-03-13
Fix: (first) task tap issue

vladimir-8:
* Moved Calendar into a separate page
* Display "Go to today" button on both map and list when other day than Today is selected


vladimir-8:
Will fix: https://github.com/coopcycle/coopcycle-app/issues/1720

PR #1757, opened 2024-04-18
Products badges redesign

Pierre-LHOSTE:
#1760 

| | Before | After |
|-|-|-|
| Light |  |  |
| Dark |  |  |


vladimir-8 (approved):
The code looks good to me. 

I only have a comment about the design. I think it has too little contrast between the categories (diets and allergens) and they looks like a gradient.

PR #1767, opened 2024-04-19
Add incident support.

r0xsh:
Fixes: #1755

### DIFF
--- a/src/redux/Courier/taskActions.js
+++ b/src/redux/Courier/taskActions.js
@@ -190,7 +190,7 @@ export function loadTasks(selectedDate, refresh = false) {
   };
 }
 
-function uploadTaskImages(task, url, state) {
+function uploadEntityImages(entity, url, state) {
   const signatures = selectSignatures(state);
   const pictures = selectPictures(state);
 
@@ -201,17 +201,17 @@ function uploadTaskImages(task, url, state) {
   }
 
   const promises = files.map(file =>
-    uploadTaskImage(state.app.httpClient, url, file, {
+    state.app.httpClient.uploadFileAsync(url, file, {
       headers: {
-        'X-Attach-To': task['@id'],
+        'X-Attach-To': entity['@id'],
       },
-    }, url),
+    }),
   );
 
   return Promise.all(promises);
 }
 
-function uploadTasksImages(tasks, url, state) {
+function uploadEntitiesImages(entities, url, state) {
   const signatures = selectSignatures(state);
   const pictures = selectPictures(state);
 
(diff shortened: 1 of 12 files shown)


src/redux/Courier/taskEntityReducer.js
```diff
@@ -81,6 +84,17 @@ function replaceItem(state, payload) {
   return state;
 }
 
+
+function updateItem(state, id, payload) {
+  const index = state.findIndex(item => item['@id'] === id);
+
+  if (index !== -1) {
+    return state.map((item, i) => (i === index ? { ...item, ...payload } : item));
```
vladimir-8, line 92: why `map`? can be just `state[index] = { ...state[index], ...payload };` ?

src/redux/Courier/taskEntityReducer.js
```diff
@@ -81,6 +84,17 @@ function replaceItem(state, payload) {
   return state;
 }
 
+
+function updateItem(state, id, payload) {
```
vladimir-8, line 88: I'd call it something like `prevItems` instead of `state`

vladimir-8 (approved):

vladimir-8:
I guess we will need to wait with merging this PR until all instances are updated? Otherwise we will not be able to release the app, as it relies on newer APIs

src/redux/Courier/taskActions.js
```diff
@@ -28,6 +28,9 @@ export const MARK_TASK_FAILED_FAILURE = 'MARK_TASK_FAILED_FAILURE';
 export const START_TASK_REQUEST = 'START_TASK_REQUEST';
 export const START_TASK_SUCCESS = 'START_TASK_SUCCESS';
 export const START_TASK_FAILURE = 'START_TASK_FAILURE';
+export const REPORT_INCIDENT_REQUEST = 'REPORT_INCIDENT_REQUEST';
```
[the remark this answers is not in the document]
r0xsh: Could be in its own PR, where we refactor all the actions.

src/redux/Courier/taskEntityReducer.js
r0xsh: For immutability, objects in javascript are passed via reference. So i should not edit the `state` variable.

r0xsh: You'd to clone the object in order to use your snippet
```javascript
const updatedItems = [...prevItems]; // Create a copy of the array
updatedItems[index] = { ...updatedItems[index], ...payload };
```

It's possible that way, but i'm more used to functional paradigm

PR #1770, opened 2024-04-23
Fix icon color in header

Pierre-LHOSTE:


src/navigation/navigators/CourierNavigator.js
```diff
@@ -58,15 +58,28 @@ const styles = StyleSheet.create({
...
     </TouchableOpacity>
   );
 };
 
 const MainStack = createStackNavigator();
 
+const headerButtons = nav => (
+  <View style={styles.buttonBar}>
+    <ButtonWithIcon
+      name="settings"
```
vladimir-8, line 74: I'd call it `icon`

vladimir-8 (approved):
Approved. Please add screenshots next time. I understand that adding `before` and `after` states could be time consuming, so it would be great if you can add the `after` state at least. It helps a lot to understand what the pull requests is about

PR #1777, opened 2024-05-01
Fix: detox setup

vladimir-8:
https://github.com/coopcycle/coopcycle-ops/issues/29

- Upgraded Detox to v20
- Updated Github Actions (launch android 9 & iOS 17 emulators; run tests; save screenshots from the test runs in artifacts)
- Added a test that launches an app and connects to a demo instance: `e2e/foodtech_first_launch.spec.js`
- Disabled existing tests (have to be fixed separately; see the list in https://github.com/coopcycle/coopcycle-ops/issues/32)


Atala:
hello

the PR is fine to me, i just have the following questions

- why have a "release" and a "debug" build for detox? shouldnt we just have "release" to simplify things?
- shall we test on more iOS / android versions? I think for Android the last critical issues were happening on specific versions (Android 7?)
- i see some comments bout disk usage problem in github runners, is that solved?

vladimir-8:
> why have a "release" and a "debug" build for detox? shouldnt we just have "release" to simplify things?

To be able to run locally faster. I did not add any new tests yet, will see how that works

vladimir-8:
> shall we test on more iOS / android versions? I think for Android the last critical issues were happening on specific versions (Android 7?)

Yes! But setting up and running emulators on Github actions is somewhat painful; each run takes at least 30 minutes and often crashes randomly. Let's see if it's usable in the current setup and then try adding more emulators

vladimir-8:
> i see some comments bout disk usage problem in github runners, is that solved?

I did not see it happening again. But the fix is somewhat hacky, I remove some dev tools that github installs and use Android 9 image, because it’s smaller
```
      - name: Free Disk Space (Ubuntu)
        uses: jlumbroso/free-disk-space@main
        with:
          # all of these default to true, but feel free to set to
          # "false" if necessary for your workflow
          android: false
          # large-packages: true
          # swap-storage: true
```

PR #1834, opened 2024-07-12
Barcode scanner

r0xsh:
Linked with: https://github.com/coopcycle/coopcycle-web/pull/4512


src/components/BarcodeCameraView.js
```diff
@@ -0,0 +1,144 @@
...
+
+  //The hide behavior is a hack to force the camera to re-render
+  //if another camera is opened on modal. If we don't do this, the
+  //camera will not render.
+  const [hide, setHide] = useState(disabled ?? false);
+
+  if (!hasPermission) {
+    return (
+      <View>
+        <Text>No camera permission</Text>
```
Atala, line 51: maybe write a more helpful message? like "Allow camera in app setting to use the barcode scanner"?

Atala (commented):
tested:
- scan an assigned task -> pop up shows unassign/ok
if i click unassign then it unassigns but I am not redirected back to the "My tasks" screen
if i click OK then nothing happens, the popup close and no information shows up in task field, i am stuck on the empty barcode screen + barcode scanning does not seem to work a second time

scanning an unassigned task -> it assigns directly to me without a popup opening, is that normal?

scanning a task assigned to another rider -> pop up opens and i can assign it to myself

so all in all here are the problems :
- what does the "ok" button means to the user in the various popups? it is more a yes/no question no? so it should be a "NO" button
- when clicking on "ok" nothing shows, nothing happens (and i think you can not scan anymore)
- after  clicking unassign/assign/assign to me then you should be redirected to the tasklist or to the map (the page you were originally on)
- how one is supposed to complete a task? when i scan a task that is assigned to me then i can only unassign, not complete. also in the UI/UX when the task is assigned to me, complete the task should be the main choice

Atala:
in the new "task complete UI" 
- I think the phone label to call directly should be next to the phone number on its right (also more easily reachable by right hand thumb there)
- how one can mark task as "DONE" without incident?

Atala:
(check yarn lint errors)

r0xsh:
Fix this navigator error
>  WARN  Found screens with the same name nested inside one another. Check:
CourierNav > CourierBarcode, CourierNav > CourierBarcode > CourierBarcode

Atala:
- [ ] on the scan page, show task packages list (so you can see quickly if you have all the needed packages) + highlight the package that was just scanned
- [ ] add the possibility to scan again (maybe action button?), as someone would like to assign himself several deliveries in a row
- [ ] change the "add a note" button color to not grey (looks disabled)
- [ ] (backend problem) : it stills assign in the wrong order :
[image](https://github.com/user-attachments/assets/46cd1a06-9e30-40ae-9634-1de948f8ede7)
- [ ] add action "start task" when scanning an already assigned drop (and force-finish the pickup at the same time ?)

to elaborate on the last comment, I still think there is a UI/UX issue as you can only scan drops but you need to validate pickups as well, for example when you prepare your tour at the warehouse. so on top of my head there are two ways to do that :
 - silently validate the pickup when you assign yourself the task (it means that the package is in your hands so you have done the pickup (!))
 - when scanning, allow to explictly validate the pickup

good job nevertheless !! go get 'em

r0xsh:
> on the scan page, show task packages list (so you can see quickly if you have all the needed packages) + highlight the package that was just scanned

Maybe a bit tricky to add, like this is note a small feature

>  add the possibility to scan again (maybe action button?), as someone would like to assign himself several deliveries in a row

That would work, it only prevent to scan the same barcode twice

Atala:
ah you dont have the task serialized so you can display the packages list ? it should come with the task no?

Atala:
ah ok
maybe just show the packages list then, and i can create an issue ?

r0xsh:
I've decided to release the barcode functionality without implementing pickup validation during dropoff scanning. There are two main reasons for this:

1. It falls outside the original scope of this feature
2. The implementation has become overly complex, particularly in the web interface. Adding this to the mobile app would be even more complicated since we'd need to handle attachments (notes, signatures, proof of delivery) and our current app state architecture isn't well-suited for this new behavior.

https://github.com/coopcycle/coopcycle-web/pull/4784#issuecomment-2538571943

r0xsh:
>  maybe just show the packages list then, and i can create an issue ?

This is more or less the case, it's handled by a modal, but maybe it worth moving this out of the modal and display it in the view instead

Issue #1807, opened 2024-06-21
Rider's TaskList not automatically refreshing correctly when coming in foreground


Atala:
It seems that the rider doesn't get the tasklist updated all the time when the app comes form the background to the foreground. ex: arriving at the restaurant. So naofood and eraman are asking their riders every time they arrive at restaurant.

Root cause may be the websocket connection being broken + no hard refresh https://github.com/coopcycle/coopcycle-ops/issues/78

We propose to do a hard refresh of assigned tasks when the app comes into the background (load from api, same as when we change tasks). to gain more robustness.

- would that be possible technically?
- would that be possible without getting to much in the rider way ?maybe background load or something, like API call without locking the screen

vladimir-8:
I think it's a good addition. If we start using RTK query in the mobile app, it already has it implemented: https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#re-fetching-on-window-focus-with-refetchonfocus

FYI:

We currently update the task list ("hard-refresh") in two cases (1) when a rider clicks on a notification in the Notification center and (2) when they click on a popup. One issue might be is that the popup disappears in a few seconds, so the task list stays outdated (if the websocket connection is also broken). I had been refactoring notifications in `auto-accept` branch and added a "hard-refresh": https://github.com/coopcycle/coopcycle-app/pull/1736/commits/931f9a2261ef20f527f0591558be19a00802b8f8 if a push notification is received in the foreground (it only works on Android, on iOS push notification is displayed in the notification center even if the app is in the foreground)

If we implemented automatic refresh maybe we can remove "hard refresh" from both the popup and PushNotificationMiddleware

Atala:
yes good for me :+1: 

i think the thing with react query is nice, but may need some refinement to work on mobile
https://stackoverflow.com/questions/73951256/react-query-refetchonwindowfocus-in-react-native-not-working
https://github.com/reduxjs/redux-toolkit/issues/3652

PR #1840, opened 2024-07-31
autorefresh tasks for a courier

vladimir-8:
Fixes: https://github.com/coopcycle/coopcycle-app/issues/1807


src/redux/setupListenersReactNative.js
```diff
@@ -0,0 +1,60 @@
...
+  dispatch,
+  { onFocus, onFocusLost, onOffline, onOnline },
+) {
+  const handleFocus = () => dispatch(onFocus());
+  const handleFocusLost = () => dispatch(onFocusLost());
+  const handleOnline = () => dispatch(onOnline());
+  const handleOffline = () => dispatch(onOffline());
+
+  const _handleAppStateChange = nextAppState => {
+    const foreground = !!(
```
Atala, line 21: double `!` is negating the negation thus "useless"?

src/redux/Courier/taskEntityReducer.js
```diff
@@ -295,6 +260,74 @@ export const tasksEntityReducer = (
...
+        ...state,
+        loadTasksFetchError: false,
+        completeTaskFetchError: false,
+        // This is the date that is selected in the UI
+        date: action.meta.arg.originalArgs.format('YYYY-MM-DD'),
+        // isFetching: true,  # don't set isFetching flag to prevent global loading spinner for rtk query requests
+      };
+
+    //using axios
+    case action.type === LOAD_TASKS_SUCCESS: {
```
Atala, line 289: maybe add an issue to deprecate so we use only the rtk query ?

src/redux/api/baseQuery.js
```diff
@@ -0,0 +1,129 @@
...
+
+    if (!user) {
+      return result;
+    }
+
+    const refreshToken = user.refreshToken;
+
+    // checking whether the mutex is locked
+    if (mutex.isLocked()) {
+      // wait until the mutex is available without locking it
```
Atala, line 77: if i understand well we could add the following comment :
`// wait for the mutex release, i.e. wait that another request that needed a fresh token as well gets the new token for us and update the state accordingly`
?

src/redux/api/slice.js
```diff
@@ -0,0 +1,46 @@
...
+    }),
+    getOrderValidate: builder.query({
+      query: nodeId => `${nodeId}/validate`,
+    }),
+    updateOrder: builder.mutation({
+      query: ({ nodeId, ...patch }) => ({
+        url: nodeId,
+        method: 'PUT',
+        body: patch,
+      }),
```
Atala, line 35: I am surprised by the functions that are not `getMyTasks`
1. they are not linked to the PR
2. they are not explictly called? or maybe they will be called on onFocus/onOnline... aso, but it seems weird to me for example for `updateOrder`?
3. we should build apiSlice differently according to the user roles? (getMytasks is for courier, generateOrders for dispatchers, updateorder for admin/restaurant owner ?

Atala (commented):
hello,

i tested the behavior locally and everything seems alright :+1: it is a huge improvment :+1: and now we have that we can add easily same behavior for restaurant part and solve that much problem

my main interrogation is the `apiSlice.endpoints`. to me you are mixing endpoints that belongs to different categories of users, and instead we should have one slice for restaurant owner, one for riders, one for dispatcher, one for admin... and combine them according to the user's role?

also can we have fine-grained control on what will be refetched `onFocus` when we use the global `setupListeners` ? (i can check myslef if you dont know)

src/redux/setupListenersReactNative.js
vladimir-8: no, it's a fancy way to convert the value to boolean

src/redux/api/slice.js
vladimir-8: It's actually a copy-paste of https://github.com/coopcycle/coopcycle-web/blob/master/js/app/api/slice.js with `getMyTasks` added. I mentioned elsewhere that, ideally, I see it in a library, but for now, I think we should try to make them compatible so it's easy to copy-paste already implemented endpoints between the web and the app

1. not directly
2. no, they are not used at the moment, but will be soon. In a timing modal, for example
3. I thought about it, but rkt query actually advises against it (see Tip): https://redux-toolkit.js.org/rtk-query/api/createApi, so I decided to stick to the standard approach

vladimir-8:
> can we have fine-grained control on what will be refetched onFocus when we use the global setupListeners ? (i can check myslef if you dont know)

Each component can define the requests that they want to get refetched. It's defined here, if that's what you are asking:
https://github.com/coopcycle/coopcycle-app/blob/ad73ae221e776a2c2e6706426c0a4751fe8179ca/src/navigation/courier/TasksPage.js#L58

src/redux/api/slice.js
Atala: 3. i am not sure we understand the same, i was thinking to generate dynamically the slice upon login so you don't load the automatic refetch for tasks for a customer for example

What would happen with the current code if a simple customer logs in for example? will it try to refetch tasks (hence 401) on focus? I can test that later

vladimir-8: No, at least as I understand. Refetch happens only in `TasksPage` component https://github.com/coopcycle/coopcycle-app/blob/ad73ae221e776a2c2e6706426c0a4751fe8179ca/src/navigation/courier/TasksPage.js#L58 Until you reach it no refetch is setup. 

`setupListenersReactNative` https://github.com/coopcycle/coopcycle-app/blob/f7667c26a9c822427f6ef30aea845baf59f7436a/src/redux/setupListenersReactNative.js only tell rtk about the state changes, but they don't trigger any requests until they are configured elsewhere

Atala:
ah ok i had a misunderstanding of the RTK doc I think, because when I read it I thought the behavior will be enabled by default for all endpoints when calling `setupListeners` https://redux-toolkit.js.org/rtk-query/api/setupListeners

maybe change your comment `enable refetchOnFocus and refetchOnReconnect behaviors` to something like `bind handlers for refetchOnfocus (need to be enabled in the component` ?

Atala (approved):

src/navigation/courier/TaskListPage.js
```diff
@@ -29,108 +27,67 @@ const styles = StyleSheet.create({
...
+
+  const containerStyle = [styles.container];
+  if (tasks.length === 0) {
+    containerStyle.push(styles.containerEmpty);
   }
 
-  render() {
-    const { tasks, tasksWithColor, selectedDate } = this.props;
+  const { isFetching, refetch } = useGetMyTasksQuery(selectedDate, {
+    refetchOnFocus: true,
```
Atala, line 45: @vladimir-8 we should enable `refetchOnReconnect` here as well no ?

vladimir-8: We can do it, why not

src/navigation/courier/TaskListPage.js
```diff
@@ -29,108 +27,67 @@ const styles = StyleSheet.create({
...
+  });
```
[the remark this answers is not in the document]
vladimir-8: As I understand, all screens with `refetchOnfocus` == true will try to fetch when they get focused, so we should be careful with using this option, otherwise we end up with a lot of requests. If  `refetchOnfocus` == false then there will be only one request and all other screens should get the data from the cache: https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#default-cache-behavior 

I think, ideally, it would be cool to set a lifetime on each item in the cache, like 60 seconds, in that case we don't need `refetchOnfocus` we will just drop the old cached data and trigger a new request. I don't see any build-in ways to do it, so it seems we need to modify https://redux-toolkit.js.org/rtk-query/api/createApi#onquerystarted method ourselves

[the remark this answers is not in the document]
vladimir-8: There is `keepUnusedDataFor` https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#reducing-subscription-time-with-keepunuseddatafor option, but I guess a component does not unsubscribe when it loses focus. If it's the case, then it will not work for us

Issue #1745, opened 2024-04-06
Setup datadog for React Native


vladimir-8:
https://docs.datadoghq.com/integrations/rum_react_native/

PR #1846, opened 2024-08-12
added: datadog

vladimir-8:
Fixes: https://github.com/coopcycle/coopcycle-app/issues/1745


src/Datadog.js
```diff
@@ -0,0 +1,135 @@
...
+}
+
+const applicationId = Config.DATADOG_APPLICATION_ID;
+if (enabled && !applicationId) {
+  throw new Error('DATADOG_APPLICATION_ID is required');
+}
+
+const datadogConfig = new DatadogProviderConfiguration(
+  clientToken,
+  __DEV__ ? 'dev' : 'prod',
```
Atala, line 27: I am not sure we want to send logs in "dev" build? what is the use case for "dev builds? beta app?

src/Datadog.js
```diff
@@ -0,0 +1,135 @@
...
+datadogConfig.site = 'US1';
+// Optional: Enable JavaScript long task collection
+datadogConfig.longTaskThresholdMs = 100;
+// Optional: enable or disable native crash reports
+datadogConfig.nativeCrashReportEnabled = true;
+// Optional: Sample RUM sessions (% of session are sent to Datadog. Default is 100%).
+datadogConfig.sessionSamplingRate = __DEV__ ? 100 : 50;
+// Optional: Sample tracing integrations for network calls between your app and your backend (% of calls to your instrumented backend are linked from the RUM view to the APM view. Default is 20%)
+// You need to specify the hosts of your backends to enable tracing with these backends
+datadogConfig.resourceTracingSamplingRate = __DEV__ ? 100 : 50;
```
Atala, line 43: 50 maybe a little bit too much we are not going to be overwhelmed you think?

Atala (commented):
hello

it is very good indeed :+1: I can not test it right now because I don't have my testing mobile phone :/

- i was wondering if we could some "tagging" of reports to get the instance name the app is connected to. like DatadogProviderConfiguration.setAttributes({coopcycle_app_name: <domain_name>})
- i am personnaly not very interested for navigation logging because I think we will get overwhelmed but i guess you want it for some very specific reason?
- in this case, 50% seems to me to much should be less in my opinion
- maybe what could be nice is that the user can have a specific way to enable 100% logging, like a hidden setting or whatever?

src/Datadog.js
vladimir-8: That's only for debugging the datadog setup during development. By default, it will be disabled and has to be enabled at https://github.com/coopcycle/coopcycle-app/pull/1846/files#diff-418d449e0322d2e2e548fcb698159345175ca7c9b097ce5ba96a159617dd37d1R13

src/Datadog.js
vladimir-8: I did not know what to put, actually. Ideally, I want 100% 😄 but they cost $ 1.50 per 1000 sessions. Let's start with 20% maybe

vladimir-8:
> i was wondering if we could some "tagging" of reports to get the instance name the app is connected to. like DatadogProviderConfiguration.setAttributes({coopcycle_app_name: <domain_name>})

That seems possible, I will add that

vladimir-8:
> maybe what could be nice is that the user can have a specific way to enable 100% logging, like a hidden setting or whatever?

Logging is always set at 100% as I understood. The sample rate is configured for sessions (user actions)

Atala:
> Logging is always set at 100% as I understood. The sample rate is configured for sessions (user actions)

yes yes, I was meaning to enable 100% logging for sessions. maybe not useful yet.

PR #1873, opened 2024-09-17
fixed: navigate to CheckEmail page if an account is not confirmed

vladimir-8:


src/redux/App/actions.js
```diff
@@ -452,35 +452,57 @@ export function setCurrentRoute(routeName) {
...
+            password: i18n.t('INVALID_USER_PASS'),
+          }),
+        );
+      } else {
+        dispatch(authenticationFailure(i18n.t('TRY_LATER')));
+      }
+      return;
+    }
+
+    if (!user.enabled) {
```
vladimir-8, line 492: Here is the fix

PR #1879, opened 2024-09-23
allow connecting to a local instance from an iPhone

vladimir-8:


vladimir-8:
Not needed at the moment. I needed again try setting up only for dev builds using something like https://www.reddit.com/r/reactnative/comments/l8vppl/solution_not_having_to_turn_onoff_app_transport/

PR #1899, opened 2024-10-10
Allow reporting unexpected number of zero waste containers as an incident

### LINKED ISSUE #1894: Report an incident when customer does not give expected amount of zero waste containers
When completing a **dropoff** task that has zero waste containers that are expected to be returned by the customer, if the number of containers is not the one we expect, the rider should be able to report a specific incident. 

This incident should trigger a call to the IMPEC HTTP API server side, to report the problem to IMPEC.

alexsegura:
Fixes #1894

Needs coopcycle/coopcycle-web#4699

When reporting this type of incident, riders will be able to specify the number of containers that have actually been returned. 

zero_waste_incident

### DIFF
--- a/src/navigation/task/Complete.js
+++ b/src/navigation/task/Complete.js
@@ -18,6 +18,7 @@ import React, { useEffect, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import {
   Dimensions,
+  FlatList,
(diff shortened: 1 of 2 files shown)


alexsegura:
I'd like to merge #1900 first, and rebase this PR on it.

PR #1903, opened 2024-10-17
Try to fix multiple printer tickets

alexsegura:


src/navigation/restaurant/components/OrderButtons.js
```diff
@@ -25,7 +25,8 @@ const Comp = ({ order, isPrinterConnected, onPrinterClick, printOrder }) => {
         {isPrinterConnected && (
           <Button
             endIcon={<Icon as={FontAwesome} name="print" size="sm" />}
-            onPress={printOrder}>
+            onPress={printOrder}
+            isDisabled={disablePrintButton}>
```
vladimir-8, line 29: To be honest, I don't see much value in passing these props through three components (Order.js > OrderHeading.js > OrderButtons.js). I would put them in `OrderButtons` directly

src/redux/Restaurant/selectors.js
```diff
@@ -244,3 +244,8 @@ export const selectIsActionable = createSelector(
     );
   },
 );
+
+export const selectIsPrinting = createSelector(
```
vladimir-8, line 248: There is a selector called `selectPrintingOrderId`. It can be used either here or directly from a React component

vladimir-8 (approved):

src/redux/Restaurant/reducers.js
```diff
@@ -567,7 +567,10 @@ export default (state = initialState, action = {}) => {
       const printTask = state.ordersToPrint[orderId];
 
       if (!printTask) {
-        return state;
+        return {
+          ...state,
+          printingOrderId: null,
```
vladimir-8, line 572: My idea was that only `printFulfilled`/`printRejected` actions with the same order ID as in `printPending` should reset the state (assuming that printing is sequential)

vladimir-8: Did you see a situation when it was not working?

PR #1913, opened 2024-11-08
Delivery form pickup

Pierre-LHOSTE:
https://github.com/coopcycle/coopcycle-app/issues/1792

| Pickup (default address) | Pickup (custom address) |
|-|-|
| Capture d’écran 2024-11-12 à 15 57 56 | Capture d’écran 2024-11-12 à 15 59 00 | 

| Dropoff | Time slot & packages |
|-|-|
| Capture d’écran 2024-11-12 à 15 58 21 | Capture d’écran 2024-11-12 à 15 59 21 |

PR #1920, opened 2024-11-20
Fix delivery select style

Pierre-LHOSTE:
Fix the misalignment and the bug in the dark mode of the AddressAutocomplete

https://github.com/coopcycle/coopcycle-app/pull/1913#discussion_r1834935417
https://github.com/coopcycle/coopcycle-app/pull/1913#discussion_r1834936621


r0xsh (approved):
The code is fine, the behavior work as expected :+1: 
UI side a small fix may be needed on the other screens using the autocomplete field
image

PR #1921, opened 2024-11-21
Fix/centrifugo reconnect

vladimir-8:


src/redux/middlewares/CentrifugoMiddleware/index.js
```diff
@@ -16,81 +15,137 @@ import {
...
-    if (!isCentrifugoAction(action)) {
       return next(action);
     }
 
-    const state = getState();
-
     if (
-      !selectIsAuthenticated(state) ||
-      !selectHttpClientHasCredentials(state)
+      action.type === disconnect.type ||
```
Atala, line 131: shouldnt it be `&&` ?

src/redux/middlewares/CentrifugoMiddleware/actions.js
```diff
@@ -1,16 +1,13 @@
...
-export const CONNECTED = '@centrifugo/CONNECTED';
-export const DISCONNECTED = '@centrifugo/DISCONNECTED';
 
-export const connect = createAction(CONNECT);
+export const connect = createAction('@centrifugo/CONNECT');
+export const disconnect = createAction('@centrifugo/DISCONNECT');
 
-export const connected = createAction(CONNECTED);
-export const disconnected = createAction(DISCONNECTED);
+export const connected = createAction('@centrifugo/CONNECTED');
```
Atala, line 9: should it be renamed `centrifugoConnected` for more clarity? (and other the same)

src/redux/middlewares/CentrifugoMiddleware/index.js
vladimir-8: No, that's correct. We want to proceed only if both are true. To be honest, I think the second check is redundant, but I don't want to experiment with it now 🙈

src/redux/middlewares/CentrifugoMiddleware/actions.js
vladimir-8: Done

src/redux/middlewares/CentrifugoMiddleware/index.js
Atala: that's why i am wondering it should be AND here? instead of OR

vladimir-8: The check is if NOT (`selectIsAuthenticated` AND `selectHttpClientHasCredentials`), which is the same as (NOT `selectIsAuthenticated`) OR (NOT `selectHttpClientHasCredentials`)

Atala: you are speaking about the removed code no?

Atala (approved):
tested OK with the following
- as restaurant, set airplane mode and the push connection comes back alright after reconnect 
- as restaurant, set lock screen and the push connection comes back alright after reconnect
- as rider, set airplane mode and the push connection comes back alright after reconnect

do you want to deploy to everybody or test in a few restaurants? we could use restaurant owner emails for that?

Issue #1864, opened 2024-09-10
When making a new order on the application as a store owner, the client field is empty even after i have picked a client


Paul-Eraman-CoopCycle:
Please add gif: @Pierre-LHOSTE

Atala:
see if fixed in #1527

Paul-Eraman-CoopCycle:
@Pierre-LHOSTE is this fixed?

PR #1938, opened 2025-01-08
Add selected address client name

Pierre-LHOSTE:
https://github.com/coopcycle/coopcycle-app/issues/1864

I don't know why but `clearButtonMode="while-editing"` doesn't work


r0xsh:
> I don't know why but `clearButtonMode="while-editing"` doesn't work
Seems like it's only supported on iOS: https://reactnative.dev/docs/textinput#clearbuttonmode-ios

r0xsh (approved):

PR #1942, opened 2025-01-22
Allow paying with Paygreen via redirect URL.

### LINKED ISSUE #1941: Implement PayGreen - APP
See https://github.com/coopcycle/coopcycle-web/issues/3042

"
What we want is to be able to use Stripe **OR** PayGreen. 
For each restaurant, we need to add a property that allows to choose between Stripe and PayGreen. 

- [x] Allow storing PayGreen app id & private key in platform settings
- [x] Allow connecting each restaurant with PayGreen
- [ ] Change `GatewayResolver::resolve` to pass the order as parameter, and decide which gateway to use based on the order context (if the restaurant of the order should use PayGreen, use it)
- [ ] Allow choosing between Stripe & Paygreen at restaurant level

See libs 
- https://paygreen.github.io/paygreen-ui-doc/docs/
- https://github.com/PayGreen/paygreen-php

See examples 
- https://github.com/RenaudGreen/PayGreenPrestashop
- https://github.com/PayGreen/sylius-paygreen-plugin

See docs 
- https://github.com/PayGreen/paygreen-php/tree/master/docs/v2
- https://paygreen.fr/documentation/api-documentation-categorie
"

alexsegura:
Fixes #1941

### DIFF
--- a/src/navigation/checkout/Payment.js
+++ b/src/navigation/checkout/Payment.js
@@ -69,7 +69,7 @@ const CreditCard = ({ cart, paymentMethods, paymentGateway, loadPaymentMethods, 
         navigation.navigate(routesByMethod[type]);
       }
     })
-  }, []);
+  }, [ navigation, paymentGateway, setPaymentMethod ]);
 
   useEffect(() => {
     loadPaymentMethods();


r0xsh (changes requested):
The feature is working as expected and should work on modern devices.
I was worried about using external browser on which we don't have any control and know which browser the user is using, i tried it with Firefox focus and still the payment work fine :+1: 
However on lower spec devices i was worried about the OOM background app killer feature present on both iOS and Android, so i tried killing CoopCycle while i was in the browser to see if the state restoration was well handled, sadly it is not and lead to an infinite loading state :/

https://github.com/user-attachments/assets/f389d007-d608-4edc-8576-e7d9b47c8ece

alexsegura:
@r0xsh I used [react-native-inappbrowser-reborn](https://github.com/proyecto26/react-native-inappbrowser), can you test again?

r0xsh (approved):

android/build.gradle
```diff
@@ -14,6 +14,9 @@ buildscript {
         googlePlayServicesLocationVersion = "21.0.1"
         mercadoPagoPxVersion = "4.53.2"
         reactNativeVersion = "0.73.5" // https://github.com/expo/expo/issues/18129
+        // https://github.com/proyecto26/react-native-inappbrowser#mostly-automatic-installation
+        androidXAnnotation = "1.2.0"
+        androidXBrowser = "1.3.0"
```
alexsegura, line 19: @r0xsh can you try on Android with latest versions of those? 
- https://mvnrepository.com/artifact/androidx.annotation/annotation
- https://mvnrepository.com/artifact/androidx.browser/browser

Issue #197, opened 2019-06-21
Fatal error when open a task with small screen


florentk:
When I open a task (push on the task in task list), the app cause an error :
```
06-21 07:07:48.900 21041 21041 E AndroidRuntime: FATAL EXCEPTION: main
06-21 07:07:48.900 21041 21041 E AndroidRuntime: Process: fr.coopcycle, PID: 21041
06-21 07:07:48.900 21041 21041 E AndroidRuntime: com.google.maps.api.android.lib6.common.apiexception.d: Error using newLatLngBounds(LatLngBounds, int): View size is too small after padding is applied.
```
The task must have a next task. 
Mobile model : Unihertz Jelly Pro
Version Android : 7.0
Screen size : 240x432 pixel (according to the manufacturer)

alexsegura:
Fixed in 1d48c49eca8c87711585042140781b781f9b4a49

PR #1949, opened 2025-02-10
[PERTE-197] Be able to create an order for a B2B customer from the mobile app

diegomanuel:
### Issues: [#197](https://github.com/coopcycle/coopcycle/issues/197) & [#246](https://github.com/coopcycle/coopcycle/issues/246)

- [x] Ensure to fetch ALL stores! Currently we fetch the max allowed by the API endpoint: 100
- [x] Add an unit test for the new function above!
- [x] Fetch all the stores only when clicking the "+" button instead of when entering the dispatch screen directly
- [x] Make sure the "back" button arrow really goes back (now it currently closes the whole new delivery flow)
- [x] Sort the stores alphabetically (done in the app for now, could be done in the backend app in the future)
- [x] Mark/rename unused files as: "DEPRECATED-....."
- [x] Rename `StoreListInput` to `StoreListSelect`
- [x] Make sure we don't hide any store from the list when the keyboard is displayed
- [x] Remove special characters from user input text and from store names prior to compare them to filter
- [x] Change "Add Task" text on the top to "New Delivery"
- [x] Complete the missing multilang tags we've introduced (ie, the "New Delivery" text mentioned above)
- [x] Remove the "Next" button at the botton since is not going to be used. It goes forward/next when clicking an item in the store list.
- [x] The item above will probably make the "Formik" form component (and "ModalFormWrapper) obsolete!
If that's the case, remove everything's not needed.
- [x] Refresh the stores FlatList (by swipping down the list)
- [x] Show the spinner while loading the stores
- [x] Make the tests to work (unit & e2e)
- [x] Remove the test file `e2e/01_dispatch.spec.js` and create a new one for the new flow
- [x] Remove the saved list of stores on logout.
NO! We don't store the stores list anymore into the state
- [x] Make sure to delete `stores` from` state.dispatch.stores`


src/navigation/delivery/NewDeliveryStore.js
vladimir-8, line 1: I guess you did not add a way to reload stores, or am I missing it?

vladimir-8, line 1: Also, we should remove the saved list of stores on logout

vladimir-8 (commented):
Nice work! Please check the comments

src/redux/Dispatch/actions.js
```diff
@@ -42,6 +43,10 @@ export const LOAD_TASK_LISTS_REQUEST = 'LOAD_TASK_LISTS_REQUEST';
 export const LOAD_TASK_LISTS_SUCCESS = 'LOAD_TASK_LISTS_SUCCESS';
 export const LOAD_TASK_LISTS_FAILURE = 'LOAD_TASK_LISTS_FAILURE';
 
+export const LOAD_STORES_REQUEST = 'LOAD_STORES_REQUEST';
```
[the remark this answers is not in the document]
diegomanuel: Sure! But the problem is that all actions already are like that so it might be better to create another issue/ticket to do all that refactor at once, is that sounding good @vladimir-8? I can create the issue/ticket.

src/navigation/delivery/NewDeliveryStore.js
diegomanuel: We aren't storing the stores list anymore!

e2e/01_dispatch.spec.js
```diff
@@ -30,15 +30,16 @@ describe.skip('Dispatch', () => {
     }
   });
 
+  // TODO flow changed #197 - needs fixing
```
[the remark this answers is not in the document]
diegomanuel: Sure! We'll delete this test case file because that flow is not used anymore and it will be replaced with the "NewDelivery" flow with a new test case file for it.

src/i18n/locales/an.json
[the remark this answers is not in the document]
diegomanuel: Good to know, thanks!

src/navigation/delivery/NewDeliveryStore.js
vladimir-8: Why did you remove it? I think 'caching' them in redux and having pull-to-refresh would be working fine

diegomanuel: We think is better to go get the store list every time the user clicks the "+" button.
We changed the behaviour, before our changes the store list was getting fetched when the user entered the dispatch screen. Now is fetched only when it clicks the "+" button.

Do you really want us to "cache" the store list in redux @vladimir-8 ?
NOTE: The application isn't cleaning anything from the state when doing a logout (just the user token). So if we "cache" the store list and you want us to clean it on logout, it'd be the first thing to be cleaned out.

vladimir-8: Well, it's not a must have, but I think it would be a better user experience. In my understanding the list of clients ("stores") do not change that often and we will be making extra requests on every single order...

> The application isn't cleaning anything from the state when doing a logout

It should. See reducers and middlewares handling `LOGOUT_SUCCESS` action

vladimir-8 (approved):

src/redux/Delivery/actions.js
```diff
@@ -15,6 +15,7 @@ export const LOAD_TIME_SLOT_SUCCESS = '@delivery/LOAD_TIME_SLOT_SUCCESS';
 export const LOAD_TIME_SLOTS_SUCCESS = '@delivery/LOAD_TIME_SLOTS_SUCCESS';
 export const SET_REFRESHING = '@delivery/SET_REFRESHING';
 export const SET_STORE = '@delivery/SET_STORE';
+export const SET_STORES = '@delivery/SET_STORES';
```
vladimir-8, line 18: Not a blocker, but I think we should get rid of these extra variables and define actions as:
`export const setStores = createAction('@delivery/SET_STORES');`

(When you need an action name inside a reducer you can use `case setStores.type:`)

src/navigation/delivery/NewDeliveryStore.js
```diff
@@ -21,13 +22,29 @@ const NewDeliveryStore = (props) => {
...
+    if (backendStores) {
+      dispatch(setStores(backendStores));
+    }
+  }, [backendStores, dispatch]);
+
+  useEffect(() => {
+    setIsLoading(isLoadingBackendStores && (
+      !stores || stores.length === 0
+    ));
+  }, [isLoadingBackendStores, stores]);
```
vladimir-8, line 47: Could it be `const isLoading = useMemo(() => isLoadingBackendStores && (!stores || stores.length === 0), [isLoadingBackendStores, stores]);` instead of useState and useEffect?

vladimir-8 (approved):

diegomanuel:
Sure! Will address both of them in our new PR/branch:
https://github.com/coopcycle/coopcycle-app/pull/1958

Issue #1945, opened 2025-02-05
As a customer, I am unable to desactivate the "zero waste" option in the app


Atala:
To reproduce :
- create an order in a restaurant where zero waste is activated
- uncheck "zero waste"
- try to click on "Order", it will still display the error message "you need to connect your Boite Nomades account"

![Image](https://github.com/user-attachments/assets/676a8f32-94d1-4752-b428-45189023c32e)

alexsegura:
I think it was related to a bug in NativeBase Checkbox when the component is controlled. 
Actually, the checkbox was "reversed", it was showing as unchecked when checked and vice-versa. 

I replaced it with a very simple component based on vector icons, and was able to verify that it works better by putting them side by side.

alexsegura:
Just to explain, I was able to reproduce the problem by clicking very quickly on the "Order" button at the bottom after having clicked on the button in the popup. If you wait a little bit after clicking on the button in the popup, so that HTTP request to update cart has finished, the checkbox works as expected. 

![Image](https://github.com/user-attachments/assets/6825af49-2368-45b5-9fc4-f7c5ff95817a)

I think the whole thing could be improved by refactoring how loading state is managed in checkout, so that the button can't be clicked until state is "stable". 
There are multiple state properties used to represent loading state (`isLoading`, `isFetching`, `loadingCarts`...), it's a little mess.

Atala:
tested not working in naofood 0.30

Issue #1824, opened 2024-06-27
As a consumer, when I make a 2ond order, the zero waste delivery is not activated by default


impec-dev:
**Describe the bug**
As a consumer, when I make a 2ond order, the zero waste delivery is not activated by default

**To Reproduce**
1. Make a zero-waste food order with a new or existing account (zero waste option is activated by default)
2. Once the order is validated and in preparation, stard a new order with the same consumer account
3. Select zero waste dish
4. See that the zero-waste order is not activated

**Expected behavior**
zero waste option must be activated by default. 

**Screenshots**

https://github.com/coopcycle/coopcycle-app/assets/83373118/da262f48-64f3-4a4e-9967-f27bf537a19e

PR #1953, opened 2025-02-17
Fix for "As a customer, I am unable to desactivate the "zero waste" option in the cart at checkout"

Atala:
Actually what happen is :
 - the customer disable zerowaste on the cart
 - the cart is updated
 - the event CHECKOUT_SUCCESS is dispatched
 - `shouldAskToEnableReusablePackaging` opens the modal
 - zero waste is activated again

#1945

the comment says that `shouldAskToEnableReusablePackaging` was used to solve #1824 but i dont reproduce the problem in this issue after removing this line. another problem is maybe the event CHECKOUT_SUCCESS because its name state it should be sent only once but it is sent several times, maybe every time the cart is updated?

### DIFF
--- a/src/redux/Checkout/reducers.js
+++ b/src/redux/Checkout/reducers.js
@@ -213,6 +213,9 @@ export default (state = initialState, action = {}) => {
           ...state.carts,
           [action.payload.cart.restaurant]: action.payload,
         },
+        // After a successful checkout, we may ask again to enable zero waste
+        // https://github.com/coopcycle/coopcycle-app/issues/1824
+        shouldAskToEnableReusablePackaging: true
       };
 
     case INIT_CART_FAILURE:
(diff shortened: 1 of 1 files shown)


src/redux/Checkout/reducers.js
```diff
@@ -297,10 +297,7 @@ export default (state = initialState, action = {}) => {
       return {
         ...state,
         isFetching: false,
-        errors: [],
-        // After a successful checkout, we may ask again to enable zero waste
-        // https://github.com/coopcycle/coopcycle-app/issues/1824
-        shouldAskToEnableReusablePackaging: true,
+        errors: []
```
vladimir-8, line 300: I don't understand how it can work without that code. It should reset that flag somewhere; otherwise, there would be no prompt next time.

If I'm to do it, I'd put inside `case INIT_SUCCESS:` instead (`CHECKOUT_SUCCESS` is actually a misleading name currently, as it actually means `ANY_CART_REQUEST_SUCCESS`)

Atala: hi

thanks for testing, i think i tested to quickly. indeed INIT_SUCCESS seems to do the trick

vladimir-8 (approved):

PR #1954, opened 2025-02-18
Use google places (new)

Atala:


src/components/AddressAutocomplete.js
```diff
@@ -119,39 +119,62 @@ function AddressAutocomplete(props) {
...
+        }
+
+        setResults(results);
+      }).catch(error => {
+        console.error('AddressAutocomplete; _autocomplete', error)
+      })
+    
     }
-  }, 300);
+  }, 400);
```
vladimir-8, line 177: I suspect there is something wrong with `bounce` logic, it sends requests suspiciously often:

Screenshot 2025-02-20 at 16 41 28

I also think we can `bounce` more aggressively to decrease the number of requests. Maybe 1 second? (Do the same on the web?)

Atala: thanks for the testing
the debounce needed obviously to be wrapped in useCallback :laughing: 
i have done some testing and 1s seems a little bit too much to me, i have set to 750ms
i will check on web

src/components/AddressAutocomplete.js
```diff
@@ -76,82 +76,116 @@ function AddressAutocomplete(props) {
...
-            }),
-          );
-
-          let results = normalizedResults.concat(normalizedPredictions);
-
-          if (normalizedResults.length > 0 && results.length > 5) {
-            results = results.slice(0, 5);
+  const fuse = useMemo(() => new Fuse(addresses, fuseOptions), [addresses]);
+
+  const autoCompleteDebounced = useMemo(
```
vladimir-8, line 81: I replaced `useCallback` with `useMemo` in https://github.com/coopcycle/coopcycle-app/pull/1954/commits/cb261ca1fa309200c14eb9d69f70bf48a2902e52 to fix this error:

Screenshot 2025-02-24 at 12 07 13

vladimir-8: It's important to correctly list all parameters a hook depends on (`[country, fuse, postcode]);` part), otherwise it will use old values leading to a weird bugs

PR #1957, opened 2025-02-21
Feat: add a way for the rider to refuse background geolocation (he will be prompted again at next login)

Atala:


src/redux/middlewares/GeolocationMiddleware/index.js
```diff
@@ -109,11 +119,12 @@ export default ({ getState, dispatch }) => {
 
           willDiscloseBackgroundPermission(
             state.app.hasDisclosedBackgroundPermission,
-          ).then(() => {
+            state.app.didAcceptBackgroundGeolocation
+          ).then((didAcceptBackgroundGeolocation) => {
             // This will be persisted in AsynStorage
-            dispatch(backgroundPermissionDisclosed());
+            dispatch(backgroundPermissionDisclosed(didAcceptBackgroundGeolocation));
```
vladimir-8, line 125: It all works, but I did not understand why `didAcceptBackgroundGeolocation` redux state was needed. It seems to me that having `hasDisclosedBackgroundPermission` is enough.

So, if `didAcceptBackgroundGeolocation` is false here (as returned from `willDiscloseBackgroundPermission`), 

- we will not call `BackgroundGeolocation.start` and 
- will not dispatch `dispatch(backgroundPermissionDisclosed());` (that leaves `hasDisclosedBackgroundPermission` set to false). That will trigger the same prompt on the next launch

Atala: yes indeed not needed, I put it there at first but then changed my mind and didnt clean
will do!

PR #1958, opened 2025-02-24
[PERTE-240] Fetch all tasks and couriers

diegomanuel:
### Issue: [#240](https://github.com/coopcycle/coopcycle/issues/240)

As the title says, we need to make sure to fetch all tasks and couriers for the dispatch screen/section.
Instead of using pagination, just fetch the whole list and store it into redux.

- [x] Make sure to correctly sort couriers alphabetically
- [x] Review the tests that seemed to be working fine before, but they are now always failing

Changes:
- defined enpoints' calls to get all pages in api slice: get task, get task lists, get courier users
- used those hooks to load tasks, task lists and couriers in Dispatch view
- change component's definitions from class to function: Date (Dispatch), UnassignedTask
- deleted unused logic in actions(Dispatch)


src/navigation/dispatch/UnassignedTasks.js
vladimir-8, line 1: I get an error while loading data on this page:

https://github.com/user-attachments/assets/29a2be67-c169-45fa-876c-00898643b322

vladimir-8, line 1: There is missing color bar indicating which tasks belong to the same order (on the left), on production it looks like this:

vladimir-8, line 1: Tasks appear stuck and not removed from the list after they are assigned (happens with both single and multiple tasks)


https://github.com/user-attachments/assets/a1b74e81-18f9-4c59-ac2b-94ba7d719e5a

src/redux/api/slice.js
```diff
@@ -18,6 +21,51 @@ export const apiSlice = createApi({
         body: {},
       }),
     }),
+    getUnassignedTasks: builder.query({
+      async queryFn(date, _queryApi, _extraOptions, fetchWithBQ) {
+        const result = await fetchAllRecordsUsingFetchWithBQ(
+          fetchWithBQ,
+          'api/tasks',
```
vladimir-8, line 28: for this endpoint, pagination has to be enabled explicitly: 
`/api/tasks?date=2024-12-01&pagination=true&itemsPerPage=2&page=1`

src/redux/api/slice.js
```diff
@@ -18,6 +21,51 @@ export const apiSlice = createApi({
...
+          });
+
+        return result ? { data: result } : { error: "result.error" };
+      },
+    }),
+    getCourierUsers: builder.query({
+      async queryFn(_arg, _queryApi, _extraOptions, fetchWithBQ) {
+        const result = await fetchAllRecordsUsingFetchWithBQ(
+          fetchWithBQ,
+          '/api/users',
```
vladimir-8, line 55: For this endpoint, pagination has to be enabled explicitly (after this PR is merged: https://github.com/coopcycle/coopcycle-web/pull/4929 also it will enable this endpoint for role Dispatcher)

vladimir-8:
Depends on API changes added in https://github.com/coopcycle/coopcycle-web/pull/4929

vladimir-8 (changes requested):

src/navigation/dispatch/UnassignedTasks.js
vladimir-8: Actually, that might be due to a way I generated test data: https://github.com/coopcycle/coopcycle-web/pull/4929/files#diff-bfec92e5de40f0f9a8246d5644b9963088e99568ee5a015966e9b5bc88e0f988R67 

Does it work for you?

src/components/TaskList.js
```diff
@@ -93,7 +93,7 @@ class TaskList extends Component {
   }
 
   render() {
-    const { refreshing, onRefresh } = this.props;
+    const { refreshing, refresh } = this.props;
```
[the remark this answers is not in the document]
JuanIgAcosta: We rolled back the changes in this file

src/navigation/dispatch/UnassignedTasks.js
```diff
@@ -1,126 +1,140 @@
...
+
+  useEffect(() => {
     InteractionManager.runAfterInteractions(() => {
-      this.props.initialize();
+      dispatch(initialize());
     });
-  }
+  }, [dispatch]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
JuanIgAcosta: We moved all fetching logic and load into state to `useLoadUnassignedTasksInfo` adn removed the therefore deprecated `useLoadAllTasks`

src/hooks/useFetchAllStores.js
```diff
@@ -4,7 +4,8 @@ import { useFetchAllRecords } from "./useFetchAllRecords";
 import { sortByName } from "../redux/util";
 
 
-export function useFetchAllStores(options = null) {
+// TODO: move this logic to api slice
```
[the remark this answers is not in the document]
JuanIgAcosta: We moved this logic to api's slice and also removed unused files

src/navigation/dispatch/UnassignedTasks.js
```diff
@@ -1,126 +1,140 @@
...
+                task,
+                unassignedTasks,
+              )
+            }
+            allowMultipleSelection={task => allowToSelect(task)}
+            multipleSelectionIcon="user"
+            onMultipleSelectionAction={selectedTasks =>
+              assignSelectedTasks(selectedTasks)
+            }
+            refresh={refetch}
```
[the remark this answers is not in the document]
JuanIgAcosta: Done!

src/navigation/dispatch/UnassignedTasks.js
JuanIgAcosta: We missed to add unassigned tasks to dispatch's state, so this feature and live reload didn't work. In the new hook `useLoadUnassignedTasksInfo` that is implemented, so both were fixed.

JuanIgAcosta: Same as above

src/redux/api/slice.js
JuanIgAcosta: We added `pagination=true` in `fetchAllRecordsUsingFetchWithBQ` so all endpoint's calls using that function now have it

src/navigation/dispatch/useLoadUnassignedTasksInfo.js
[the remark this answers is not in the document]
JuanIgAcosta: We updated it as `useAllTasks`. As you said, is more accurate

src/navigation/dispatch/UnassignedTasks.js
diegomanuel: Solved!

src/navigation/dispatch/useAllTasks.js
```diff
@@ -1,21 +1,22 @@
...
   useGetTaskListsQuery,
   useGetUnassignedTasksQuery,
 } from "../../redux/api/slice";
 
 
-export function useLoadUnassignedTasksInfo(date) {
+export function useAllTasks(date) {
 
   const dispatch = useDispatch();
+  const [isFirstLoad, setIsFirstLoad] = useState(true)
```
vladimir-8, line 19: I don't understand why we need to manage it manually

I think we can use `isFirstLoad = isLoadingTaskLists || isLoadingUnassignedTasks || isLoadingCourierUsers`

because 

> isLoading - When true, indicates that the query is currently loading for the first time, and has no data yet. This will be true for the first request fired off, but not for subsequent requests

from https://redux-toolkit.js.org/rtk-query/usage/queries#frequently-used-query-hook-return-values

src/navigation/dispatch/useAllTasks.js
```diff
@@ -62,5 +53,7 @@ export function useLoadUnassignedTasksInfo(date) {
       refetchUnassignedTasks();
       refetchTaskLists();
     },
```
vladimir-8, line 55: `isFetching: isLoadingTaskLists || isLoadingUnassignedTasks || isLoadingCourierUsers,`

Notice that `isLoading` is different from `isFetching`

> isLoading - When true, indicates that the query is currently loading for the first time, and has no data yet. This will be true for the first request fired off, but not for subsequent requests.

> isFetching - When true, indicates that the query is currently fetching, but might have data from an earlier request. This will be true for both the first request fired off, as well as subsequent requests.

https://redux-toolkit.js.org/rtk-query/usage/queries#frequently-used-query-hook-return-values

Let's try not to mix them up

src/navigation/dispatch/UnassignedTasks.js
vladimir-8: Does not work for me. See successful PUT request in the logs, but the progress indicator does not disappear

https://github.com/user-attachments/assets/94bb5b66-6d41-4f30-aabb-6090a8872a90

vladimir-8, line 1: Also, I can still reproduce: `On the first launch, the loading indicator disappears too early, before all requests are fulfilled`


https://github.com/user-attachments/assets/fc147346-9f87-411f-9bfa-f6ec4439a7c1

vladimir-8 (changes requested):

src/navigation/dispatch/useAllTasks.js
diegomanuel: Right! That mix was solved at commit https://github.com/coopcycle/coopcycle-app/pull/1958/commits/d1a9b7602eb592869d962d41ce98f70ca496fd3b

src/navigation/dispatch/UnassignedTasks.js
diegomanuel: With our latest changes, I just could reproduce it once and the rest of the team not even once!
I've tried LOTS of times to reproduce it again but with no luck.
Plz try again with our latest changes and if it happens again, let us know with more detail about how the previous situation/state was.

diegomanuel: That should be solved now.

vladimir-8 (changes requested):
The blocking spinner issue seems solved 🎉 

One (I hope the last 😄 ) issue: I assigned a task (1020), but it stays in the Unassigned list. It should be removed from this list and be moved into the list of a messenger. (The same happens when I assign multiple tasks)

https://github.com/user-attachments/assets/88f7b9a5-4852-4a14-a9f9-cdf592c11592

src/navigation/dispatch/useAllTasks.js
```diff
@@ -0,0 +1,83 @@
...
+  const isLoading = useMemo(() => {
+    return isLoadingTaskLists || isLoadingUnassignedTasks || isLoadingCourierUsers;
+  }, [isLoadingTaskLists, isLoadingUnassignedTasks, isLoadingCourierUsers]);
+
+  const isFetching = useMemo(() => {
+    return isFetchingUnassignedTasks || isFetchingTaskLists || isFetchingCourierUsers;
+  }, [isFetchingUnassignedTasks, isFetchingTaskLists, isFetchingCourierUsers]);
+
+  useEffect(() => {
+    if (!isFetching && !isError && unassignedTasks && taskLists && courierUsers) {
```
vladimir-8, line 58: Also noticed that inside `loadUnassignedTasksSuccess` and `loadTaskListsSuccess` we append tasks and tasklists to the existing state, while we probably should remove the existing state, as we load all data from scratch at this point (maybe you have it planned in https://github.com/coopcycle/coopcycle/issues/235 ?)

diegomanuel: Thanks Vova! :raised_hands: 
We'll tackle those 2 last issues and let u know again :sweat_smile:

JuanIgAcosta: @vladimir-8 I'm seeing we have a problem with the approach we took. Dispatch view is only visible for Admins and Dispatchers, but the messages the view is expected to recibe vi web sockets to work correctly are only sent to admins. I think we should address the expected behavior for Dispatchers with [feature-235](https://github.com/coopcycle/coopcycle-app/pull/1968) as is commented in the code. Of course we could work on a walkaround for this issue in this PR but I think it will take the same amount of time as to finish 235.

Please tell us what you think

imagen

About clean state and load what endpoints response, it is already working like that: the trigger for all task's endpoint's fetches is `CHANGE_DATE` action and in all states related to tasks it returns `initialState`, all before fetches resolve.

vladimir-8: > Dispatch view is only visible for Admins and Dispatchers, but the messages the view is expected to recibe via web sockets to work correctly are only sent to admins. I think we should address the expected behavior for Dispatchers with https://github.com/coopcycle/coopcycle-app/pull/1968 as is commented in the code. Of course we could work on a walkaround for this issue in this PR but I think it will take the same amount of time as to finish 235.

Yes, let's tackle it in another issue. I found one more reason that might cause this bug:

- https://github.com/coopcycle/coopcycle/issues/306
- https://github.com/coopcycle/coopcycle/issues/305

vladimir-8: > About clean the state and load what endpoints response, it is already working like that: the trigger for all task's endpoint's fetches is CHANGE_DATE action and in all states related to tasks it returns initialState, all before fetches resolve.

I'm talking about a case where it's called by `refetch`, for example, after pull-to-refresh. In that case, it appends to the previous state.

JuanIgAcosta: @vladimir-8 We checked these scenarios:
- when a task is assigned via web
- when a task is unassigned via web
- when a task is cancelled via web
- when a task is assigned via app
- when a task is unassigned via app

As far as we know, those are the corner cases of current implementation. In all of them, the upsert is the right operation to perform in the state as it updates all loaded instances and add the missing ones. Also, becouse of how are tasks loaded it's imposible now to use `setAll

vladimir-8: @JuanchoAcosta Please test what happens if you reschedule a task or move it to another day. Does it disappear from a list in that case?

Screenshot 2025-04-03 at 11 33 43

JuanIgAcosta: Thanks! We didn't find a way to remove a task for a day. Excellent, we'll take a look into it

diegomanuel: The reschedule is working but for now the app user needs to refetch!

vladimir-8 (approved):

src/navigation/dispatch/UnassignedTasks.js
diegomanuel: Finally solved! :tada:

PR #1964, opened 2025-03-03
fixed: menu scroll offset

vladimir-8:


Atala (approved):
Tested OK on demo restaurants

Issue #1952, opened 2025-02-13
As to use the same endpoints as the web dispatch, refactor modification of TaskList object in the app to use /api/task_lists//task_lists/set_items/{date}/{username}


Atala:
For now the app use the following endpoint/classes for dispatch in order to assign tasks:
- /tasks/{id}/assign ( src/Action/Task/Assign.php )
- /tasks/{id}/unassign ( src/Action/Task/UnAssign.php )
- /tasks/assign ( src/Action/Task/BulkAssign.php )

we should stop to use these endpoints to use /api/task_lists//task_lists/set_items/{date}/{username} endpoint :
 - is used by web dispatch
 - supports tours
 - supports reordering

and then we can delete the endpoints above

diegomanuel:
Will be done as part of PR: https://github.com/coopcycle/coopcycle-app/pull/1975

PR #1968, opened 2025-03-17
[PERTE-235] Migrate to new v2 endpoints+events for tasks, task_list and tours

JuanIgAcosta:
### Issue: [#235](https://github.com/coopcycle/coopcycle/issues/235)

Relates to PR: https://github.com/coopcycle/coopcycle-web/pull/4944

---

Start using the new enpoints described in the issue.

- [x] Change LOAD_UNASSIGNED_TASKS_SUCCESS with LOAD_TASKS_SUCCESS and get all tasks (assigned and unassigned) from backend (delete parameter "assigned" from request's params)
- [x] Load tasks, task lists and tours into states 
- [x] Select a task list with related tasks (those that are directly related and also through tours)
- [x] Update tasks, task lists and/or tours using web sockets (PR [#4944](https://github.com/coopcycle/coopcycle-web/pull/4944))
- [x] Show only selected date's tasks, task lists and tours
- [x] Update reducers tests
- [x] Filter events for the current date only
- [x] Once coopcycle/coopcycle-app#1958 is merged, bring `master` here and test that everything works just fine! :sweat_smile: 
- [x] Make the order of tasks from a tour match between the web and the mobile app
- [x] When task/s from a tour is/are assigned from app, remove the task/s from the tour by calling `PUT /api/tours/{id}` to replace it with the tour's remaining tasks (even if it's left empty).
- [x] Replace `import { createAction } from 'redux-actions';` by `import { createAction } from '@reduxjs/toolkit'` just at file: [redux/Dispatch/actions.js](https://github.com/coopcycle/coopcycle-app/blob/abd5c5fe265af37ee3be08b80119ce23b0b0e550/src/redux/Dispatch/actions.js)
- [x] Make sure that unassign from app works as expected
  * Handle error response cases (the app hangs otherwise)
  * Update tours (unassing if the tasks that are going to be unassigned belongs to a tour and no tasks left in the tour)
- [x] Make sure to empty the tasks, tasklist and tours from store when the user logout/in?


vladimir-8:
I did a quick test and noticed a few small issues around tours:

1. When a task from a tour is assigned it appears in both an assigned list and a tour. I guess the simplest would be to remove a task from a tour in that case, as we currently don't have tour management implemented in the mobile app:

Screenshot 2025-04-07 at 13 19 14

2. When a tour is assigned to messenger, the order of tasks doesn't match between the web and the mobile app:

Screenshot 2025-04-07 at 13 18 55

vladimir-8:
https://github.com/coopcycle/coopcycle-app/blob/abd5c5fe265af37ee3be08b80119ce23b0b0e550/src/redux/Dispatch/actions.js

As you are actively working on this file, it would be nice to start moving away from using `createAction` from 'redux-actions' to '@reduxjs/toolkit' and cut some boilerplate in action creation:

```
import { createAction } from 'redux-actions';
```

```
import { createAction } from '@reduxjs/toolkit'
```

Ideally, it could be just a drop-in replacement, but as I'm not sure of all the ways how it's used, I took a more cautious approach in https://github.com/coopcycle/coopcycle-app/blob/master/src/redux/Checkout/actions.js

1. Rename: createAction -> createFsAction (Flux Standard Action: https://github.com/redux-utilities/redux-actions)
2. Use `createAction` from '@reduxjs/toolkit' for all new/modified actions (no need for a separate string constant): 

```
export const setPersistedTimeRange = createAction('SET_PERSISTED_TIME_RANGE');
```

diegomanuel:
Cool thankssss we'll do that adjustments!

diegomanuel:
Cool thanks @vladimir-8 we'll do that adjustments!

When you say:
> When a task from a tour is assigned it appears in both an assigned list and a tour. I guess the simplest would be to remove a task from a tour in that case, as we currently don't have tour management implemented in the mobile app.

Plz confirm that you want us to "manually" remove the task/s from the tour by calling the corresponding endpoint/s:
* If the tour has remaining tasks after assigning to the user: `PUT /api/tours/{id}` to replace it with its remaining tasks
* If the tour has not any remaining task: `DELETE /api/tours/{id}`

vladimir-8:
@diegomanuel Could you please check how it's done on the web? Let's handle it similarly

diegomanuel:
Yep we did that and we saw it's using the new endpoint described there:
https://github.com/coopcycle/coopcycle-app/issues/1952
But the tour part, we'll handle it the same as web! Just calls `PUT /api/tours/{id}` to replace it with given tasks, even if the tour is empty.

src/redux/Dispatch/actions.js
```diff
@@ -309,3 +325,34 @@ export function updateTask(action, task) {
...
+      }
+    }
+  }
+}
+
+export function updateTour(action, tour) {
+  return function (dispatch, getState) {
+    let date = selectSelectedDate(getState());
+
+    if (isSameDateTour(tour, date)) {
```
Atala, line 347: :100:

src/redux/Dispatch/actions.js
```diff
@@ -309,3 +325,34 @@ export function updateTask(action, task) {
...
   };
 }
+
+export function updateTaskList(action, taskList) {
+  return function (dispatch, getState) {
+    let date = selectSelectedDate(getState());
+
+    if (isSameDateTaskList(taskList, date)) {
+      switch (action) {
+        case 'v2:task_list:updated':
```
Atala, line 335: I dont think it is necessary to have another `case` here as you have it already here https://github.com/coopcycle/coopcycle-app/pull/1968/files#diff-24dad85d996b9bee06ecd189b21c03f6fc11b00f28794cc7fcdc84942ae3eeeeR30

Atala (changes requested):
hello i noticed two problems
1. problem when an empty tour is inserted into the tasklist -> the thing goes mad
2. seems flaky when reordering tasks into a tasklist, sometimes update, sometimes not

see GIFs below

sync-app-and-web


flaky-reordering

src/shared/src/logistics/redux/taskListUtils.js
```diff
@@ -1,13 +1,12 @@
 import _ from 'lodash';
 import moment from 'moment';
-import { tasksToIds } from './taskUtils';
 
 export function replaceTasksWithIds(taskList) {
```
Atala, line 4: I guess the name `replaceTasksWithIds` is not appropriate anymore?

src/redux/Courier/taskEntityReducer.js
```diff
@@ -338,6 +338,7 @@ const processWsMsg = (state, action) => {
     const { name, data } = action.payload;
 
     switch (name) {
+      // TODO: update to v2
```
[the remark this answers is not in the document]
diegomanuel: Nopes, the v2 update will be done in this coming issue:
https://github.com/coopcycle/coopcycle/issues/313

src/shared/src/logistics/redux/taskListUtils.js
diegomanuel: Ok will be renamed to `replaceItemsWithItemIds`.

diegomanuel:
> hello i noticed two problems
Yep we saw that happening from time to time. If you see below the message "Encountered two children with the same key" means we have duplicated items in the list and it starts to behave weirdly.
It should be solved with our latests adjustments.

Atala:
hello i managed to break it, see here it displays some tasks several times (and no "delivery color" on the left i dont know why)

image

Atala:
I have this warning as well when it updates can you fix it ?

image

diegomanuel:
> * ry color" disapearring in the list of tasks

How to reproduce?!
We never managed to get that state in where the web shows 5 tasks and the app none.
The same for the colors.
* Are you using your local backend or demo?
* Could you record the screen instead of gif/jpgs?

diegomanuel:
> I have this warning as well when it updates can you fix it ?

In which screen? We saw that problem from time to time already happening from the beginning, but it was always harmless.
Anyways, everything is going to be changed/updated with the new UI!

PR #1982, opened 2025-04-18
bump RN v0.76.9

r0xsh:
Android:
- :heavy_check_mark: build
- :heavy_check_mark: Manual tests seems to works
- :heavy_check_mark: CI tests

iOS:

- :x: build
```
[!] Unable to find a target named `CoopCycle` in project `CoopCycle.xcodeproj`, did find `CoopCycleApp` and `CoopCycleAppTests`.
```


ios/CoopCycle.xcodeproj/project.pbxproj
```diff
@@ -135,55 +101,44 @@
...
-				A5763258E52216DBA45B1C3D /* Pods_CoopCycle_CoopCycleTests.framework */,
 			);
-			name = Frameworks;
+			name = Libraries;
 			sourceTree = "<group>";
 		};
 		83CBB9F61A601CBA00E9B192 = {
 			isa = PBXGroup;
 			children = (
-				845246CE2344832500A3A1B2 /* GoogleService-Info.plist */,
```
alexsegura, line 174: This causes an error: 

> Could not get GOOGLE_APP_ID in Google Services file from build environment

Not sure why/how this was removed, but it breaks the build

ios/CoopCycle.xcodeproj/project.pbxproj
```diff
@@ -135,55 +101,44 @@
...
-				FA2E390B2225DD32002BBCB4 /* Empty.swift */,
-				FAC644352166467B00D1DA52 /* misstickle__indian_bell_chime.wav */,
```
alexsegura, line 176: This will cause the famous "ding" to fail when receiving notifications.

alexsegura (changes requested):
All input fields are losing focus on iOS
- https://github.com/expo/expo/issues/32907
- https://github.com/GeekyAnts/NativeBase/issues/5821#issuecomment-2501851527
- https://github.com/GeekyAnts/NativeBase/issues/5821#issuecomment-2504746885

alexsegura (changes requested):
I think this should be taken care of

Capture d’écran 2025-04-29 à 11 04 29

r0xsh:
> I think this should be taken care of

Yep ! But its quit easy to use the new default props pattern. But still working on RN76

alexsegura:
The app crashes on iOS after payment. I tracked the actual bug, and it's because of Stripe `CardField` (stripe/stripe-react-native#1732)

Capture d’écran 2025-05-14 à 21 27 46

---

It's fixed in version 0.45 of `@stripe/stripe-react-native`, but the problem then is that it's not compatible with Expo 52 (stripe/stripe-react-native#1923)

alexsegura (approved):
I think it's good enough

PR #1989, opened 2025-05-05
[PERTE-333] Use logistics actions on task view

JuanIgAcosta:
### Issue: [#333](https://github.com/coopcycle/coopcycle/issues/333)

Related issue: https://github.com/coopcycle/coopcycle/issues/313
Related PR: https://github.com/coopcycle/coopcycle-app/pull/1983

---
The problem arises because both `dispatcher` and `courier` screens/sections uses the same "task detail view".
Starting the task from there, when the backend responds "OK", we update the redux state from `courier` instead of `logistics`.

**Tasks:**
- [x] Starting the task should update redux state from `logistics` and/or `courier`.
- [x] Review if there is no other remaining action/call with this kind of problem at any other shared component between screens/sections.
- [x] Make an e2e test for "start task" and "mark done" actions, making sure the redux state has the right updated data.


diegomanuel (approved):
LGTM!

PR #1992, opened 2025-05-15
Little tech debt

diegomanuel:
These are some changes and additions we did to fix some small things and make our lives easier.


e2e/support/commands.js
```diff
@@ -3,17 +3,15 @@ import axios from 'axios';
 const execSync = require('child_process').execSync;
 const os = require('os');
 
-//Make sure to have the correct path to the coopcycle-web repository while running locally
-export const COMMAND_PREFIX =
-  'cd ../coopcycle-web-repo && docker compose exec -T php';
+// Make sure to have the correct path to the coopcycle-web repository while running locally!
+// You can set `COOPCYCLE_WEB_REPO` at your `.env` file and use Makefile's targets to run the tests
+const COOPCYCLE_WEB_REPO = process.env.COOPCYCLE_WEB_REPO || 'coopcycle-web-repo';
```
alexsegura, line 8: `coopcycle-web-repo` should be `coopcycle-web`

diegomanuel: Yep! But that was the previously _hardcoded_ path and just in case I didn't want to change it (it may break CI?), that's why I kept it as default value now and it can be overriden.

@alexsegura you want me to change that default to `coopcycle-web`, then?
No problem for me, that way I can remove my override at my `.env` :raised_hands:

diegomanuel: Changed at commit: https://github.com/coopcycle/coopcycle-app/pull/1992/commits/df652b0d6f1fe2b911ff0a6cb1b3a0adcb7f02cc
If nothing breaks, I'll let `coopcycle-web` then!

alexsegura: Ha ok sorry, I didn't have in mind it was the checked out repo in CI, I thought it was the local repo 🫤

diegomanuel: No problem!
I finally did the change and updated the path from `../coopcycle-web-repo` to `../coopcycle-web`  and adjusted all the needed places.
It should be now ready to be merged :crossed_fingers:

alexsegura (approved):

PR #1985, opened 2025-04-22
[PERTE-233] Refactor dispatch screen as a single list

JuanIgAcosta:
### Issue: [#233](https://github.com/coopcycle/coopcycle/issues/233)

Related issues:
* [coopcycle/coopcycle#234](https://github.com/coopcycle/coopcycle/issues/234)
* [coopcycle/coopcycle#198](https://github.com/coopcycle/coopcycle/issues/198)
* [coopcycle/coopcycle#330](https://github.com/coopcycle/coopcycle/issues/330)
* [coopcycle/coopcycle#331](https://github.com/coopcycle/coopcycle/issues/331)

Related PRs:
* https://github.com/coopcycle/coopcycle-web/pull/4965
* https://github.com/coopcycle/coopcycle-app/pull/1989

---
Follow the guidelines in the issue description to make the first basic and usable UI version for dispatcher screen.

:loudspeaker:  Take into account the desired final result, but **do not** implement the new design just yet!
We have 2 specific issues for that: [#330](https://github.com/coopcycle/coopcycle/issues/330) and [#331](https://github.com/coopcycle/coopcycle/issues/331)

image

**Tasks:**
- [x] Think about design (paddings, components, etc)
Done by @luanacuello (see [figma](https://www.figma.com/design/UId2G6Eh6LvqBDYWqpceL5/Coopcycle?node-id=143-1314&t=fSgpCK7Y4TgbjIfJ-0))
- [x] Create the issue to update needed components for task card with the new layout.
  - https://github.com/coopcycle/coopcycle/issues/330
  - https://github.com/coopcycle/coopcycle/issues/331
- [x] Check if "sticky header" from SectionList works fine.
- [x] Ensure user can assign unassigned tasks
- [x] Ensure user can unassign assigned tasks
- [x] Add refetch functionallity
- [x] Add loading spinner
- [x] Remove unused components/screens
- [x] Merge changes from PR: https://github.com/coopcycle/coopcycle-app/pull/1983
- [x] Merge changes from PR: https://github.com/coopcycle/coopcycle-app/pull/1989
- [x] [#234](https://github.com/coopcycle/coopcycle/issues/234) Add support for collapsable sections (new orders + couriers), but just to be toggleable
(do not include the arrow, that will be done at [#331](https://github.com/coopcycle/coopcycle/issues/331))
- [x] Check if everything works fine from courier section!
- [x] Add e2e tests for assign, bulk assign and unassing.

PR #1994, opened 2025-05-16
[PERTE-331] Update dispatch screen with new design and behaviour

diegomanuel:
### Issue: [#331](https://github.com/coopcycle/coopcycle/issues/331)

Related issues:
* [#233](https://github.com/coopcycle/coopcycle/issues/233) with PR [#1985](https://github.com/coopcycle/coopcycle-app/pull/1985)
* [#330](https://github.com/coopcycle/coopcycle/issues/330) with PR [#1993](https://github.com/coopcycle/coopcycle-app/pull/1993)

---

**Consider the layout to be like:**
Image

**Tasks:**
- [x] Apply the new figma design to the whole screen layout.
  - Note that we aren't including the bottom horizontal bar with the map button, etc.
  Just let the layout ready to easily add that horizontal bar later, at [#349](https://github.com/coopcycle/coopcycle/issues/349).
- [x] Make the bulk (un/re)assign icon to appear/hide and to work then pressed.
  - [x] Should support orders/tasks from different task lists.
  - [x] Floating fixed at the bottom of the screen.
  - [x] Do not remove it from the `TaskList` since it's used from courier screen!
  - [x] Handle change assignation for tasks and orders
- [x] Add an animation for section's toggle and the arrow (relates to [#234](https://github.com/coopcycle/coopcycle/issues/234))
  - [x] Make sure we keep the swiped orders/tasks when collapsing/expanding sections.
- [x] Fix the "glitch" (spinner goes away and the screen is avail to user) between unassing and re/assign when using bulk.
- [x] Check if everything works fine from courier section!
  - It shouldn't change anything there (it will for [#330](https://github.com/coopcycle/coopcycle/issues/330))
- [x] Add a few e2e tests that verifires general screen actions, like:
  - Hidding/showing a section ([#234](https://github.com/coopcycle/coopcycle/issues/234))
  - Adding/removing items from a tasklist impacts on the corresponding section, etc.
  - Swipes+actions in the different sections (unassigned and in a courier)


src/navigation/dispatch/components/BulkEditTasksFloatingButton.js
```diff
@@ -1,50 +1,80 @@
...
 
   const allSelectedTasks = useMemo(() => {
-    return _.flatMap(Object.values(selectedTasks));
+    const ordersByTaskList = selectedTasks.orders || {};
+    const tasksByTaskList = selectedTasks.tasks || {};
+
+    const orders = _.flatMap(Object.values(ordersByTaskList));
+    const tasks = _.flatMap(Object.values(tasksByTaskList));
+
+    return [...orders, ...tasks];
```
diegomanuel, line 77: I'm not sure if `selectedTasks.orders` already has all the order's `taskIds`, but we need to make sure we only have all the taskIds from the order/s, merge with `tasks` and remove possible duplicates!

PR #2005, opened 2025-06-17
[PERTE-383] Adjustments for the new dispatch screen

warmijusti:
### Tasks:
- [ ] Make the bulk assign button to correctly dis/appear after:
  - Taping and selecting unassign/any courier, if the process (spinner loading) takes several seconds, sometimes the bulk button doesn't dissapear.
  - Collapsing and expanding the sections seems to NOT affect the internal's button state (relates to the task below).
- [ ] When swipping an order, we should also swipe all the related tasks so the user can visually see that is affecting all those tasks.
  - We should not display the bulk assign button in those cases, but only if the user selects another order or a single task for other order.
- [ ] Currently, when collapsing a section, the `TaskList` is not parsed and it causes to lose the swiped state of a task when collapsing and then expanding again.
  - This was done to improve parsing/responsiveness performance, but it's far from ideal.
  - The bulk assign button seems to behave fine anyways (but review and make sure it _really_ works fine).
  - We might revert the change that avoids `TaskList` parsing and that's it for this issue now..!
  - The performance improvement thing will be tacked at [#384](https://github.com/coopcycle/coopcycle/issues/384). 
- [x] The on-screen keyboard when writing in the search text input makes the whole screen to go up, hiding the top bars.
- [ ] Add autocomplete support for "Filter by keywords +" that searches amongst resto/store, couriers and tags.
- [x] Fix the app display in "dark mode".
- [x] Fix the map that explodes in the android emulator.
- [x] Fix the task navigation when inside a task detail (it's broken).
- [ ] Try to fix the e2e test `e2e/dispatch/success__create_delivery.spec.js`
  - It's the same as `e2e/store/success__create_delivery.spec.js` but somehow the dispatch one isn't working.
  - It seems to hang in the dropoff address text input (the text is not entered in the field, it just keeps blank/empty).
  - It seems to work fine locally when runned isolated `(make e2e-android-only TESTFILE=dispatch/success__create_delivery.spec.js)` but it fails at CI.
    - Locally, this warning/error needs to be dismissed quickly because the test is not able to tap the "Next" button otherwise: `Warning: Unknown: Support for defaultProps will be removed from memo components in a future major release. Use JavaScript default parameters instead.`
      - Note that it says "from memo components", there is another similar warning/error with other than "memo" message.
    - If we can get rid of all the "Support for defaultProps will be removed.." the better..!

### DIFF
--- a/src/components/TasksMapView.js
+++ b/src/components/TasksMapView.js
@@ -300,7 +300,7 @@ class TasksMapView extends Component {
         {this.state.mapHeight && this.state.mapHeight > 0 ? (
           <ClusteredMapView
             data={data}
-            style={[styles.map, { marginBottom: this.state.marginBottom }]}
+            style={[styles.map, { marginBottom: this.state.marginBottom, flex: 1, minHeight: '100%', widht: '100%' }]}
             width={width}
             height={this.state.mapHeight}
             initialRegion={this.initialRegion}
--- a/src/navigation/dispatch/AllTasks.js
+++ b/src/navigation/dispatch/AllTasks.js
@@ -1,34 +1,27 @@
 import { View } from 'native-base';
-import { useEffect } from "react";
-import { useTranslation } from "react-i18next";
-import { ActivityIndicator, InteractionManager } from "react-native";
-import { useDispatch, useSelector } from "react-redux";
+import { useEffect } from 'react';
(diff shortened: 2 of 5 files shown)


diegomanuel:
We will continue working on this branch!

PR #2007, opened 2025-06-18
added: error handling to redux api slice

vladimir-8:


src/redux/api/slice.js
```diff
@@ -108,23 +109,24 @@ export const apiSlice = createApi({
...
     getStores: builder.query({
       async queryFn(_arg, _queryApi, _extraOptions, fetchWithBQ) {
-        const result = await fetchAllRecordsUsingFetchWithBQ(
+        return await fetchAllRecordsUsingFetchWithBQ(
           fetchWithBQ,
           '/api/stores',
           100,
         );
-
-        return result ? { data: sortByName(result) } : { error: "result.error" };
```
diegomanuel, line 126: We need to apply `sortByName(result)` before returning.

src/redux/api/slice.js
```diff
@@ -108,23 +109,24 @@ export const apiSlice = createApi({
...
           100,
           {
-            'roles[]': 'ROLE_COURIER'
-          }
+            'roles[]': 'ROLE_COURIER',
+          },
         );
 
-        return result ? { data: sortByString(result, 'username') } : { error: "result.error" };
+        if (result.error) {
```
diegomanuel, line 116: I still prefer the 1line ternary `if` with only 1 `return`

src/redux/util.js
```diff
@@ -38,40 +38,75 @@ export function sortByString(list, key) {
...
+    } catch (err) {
+      return { data: null, error: err };
+    }
   };
-  const firstRs = await fetch(1);
 
-  if (!Object.hasOwn(firstRs, 'hydra:totalItems') || firstRs['hydra:totalItems'] <= firstRs['hydra:member'].length) {
+  const firstPageResult = await fetch(1);
+  if (firstPageResult.error) {
+    return { error: firstPageResult.error };
```
diegomanuel, line 69: Better to return the whole object: `return firstPageResult`.
We might need/want to access/use other response properties from outside (we don't know at this scope what the "outside-world" would do/need from a failed result, although we are only using this from `src/redux/api/slice.js` right now).

src/redux/util.js
```diff
@@ -38,40 +38,75 @@ export function sortByString(list, key) {
...
+    (totalItems % itemsPerPage === 0 ? 0 : 1);
+
+  return Promise.all(
+    [...Array(maxPage + 1).keys()].slice(2).map(page => fetch(page)),
+  )
+    .then(results => {
+      // Check if any page has an error
+      const errorResult = results.find(result => result.error);
+      if (errorResult) {
+        return { error: errorResult.error };
```
diegomanuel, line 96: Same here, I would just `return errorResult`;

src/redux/util.js
```diff
@@ -38,40 +38,75 @@ export function sortByString(list, key) {
...
+
+      // Combine all data from successful results
+      const combinedData = results.reduce(
+        (acc, result) => acc.concat(result.data['hydra:member']),
+        firstPageData['hydra:member'],
+      );
+
+      return { data: combinedData };
+    })
+    .catch(error => {
```
diegomanuel, line 107: Warning, just to take into account: We are changing the behaviour here.

The returned promise it's always resolved/truty because of this last `catch`.
The "outside world" would need to only depend on knowing that an "error" happened because the object `{error: ...}` is returned (instead of a rejected promise or an `await` exception).

diegomanuel (changes requested):
LGTM, just one really needed adjustment for `getStores` with a missing `sortByName(result)`.

src/redux/api/slice.js
vladimir-8: done

src/redux/api/slice.js
vladimir-8: I applied the latest prettier style, let's stick to it

PR #2025, opened 2025-08-20
[PERTE-456] Apply order detail screen new design using gluestack v3

warmijusti:
### Related PR: https://github.com/coopcycle/coopcycle-app/pull/2029

This pull request introduces Gluestack UI into the project. We are also implementing proper typing, following the recent addition of TypeScript to the project.

Image

### Tasks
- [x] Install Gluestack UI dependencies.
- [x] Create a dual UI provider.
- [x] Create a Gluestack theme.
- [x] Centralize Gluestack imports in a dedicated gluestack directory.
- [x] Centralize used types in the src/types directory.
- [x] Create AccordionHeader and AccordionContent components.
- [x] Research if we can use font-awesome icons decoupled from native-base.
- [x] Add an Assign button.
- [x] Add an assign icon button to the screen header.
- [x] Update the order detail component.
- [x] For missing data (like total order distance, order price, payment method, whatever) only display them if there is any data to be displayed.
  * Then we can add that data to be returned by the backend (inside `metadata` property) and it should automatically work on app.
  * We have a PR for adding this data: https://github.com/coopcycle/coopcycle/issues/496 
- [x] Make sure everything looks good in dark mode.
  * Map markers should be well displayed (this is a bug reported [here](https://coopcycle.slack.com/archives/C08D4U69YSV/p1755677045682759?thread_ts=1754511485.110599&cid=C08D4U69YSV))
- [x] Add payment method and the currency symbol.
- [x] Task titles for the accordeons should be: "`addr.name` - `addr.contactName`".
  * Take into account that any of the 2 could be empty, so only use the " - " sepparator if we have both.
  - [x] If a PU/DO just have an address (without name) fallback to "Pickup"/"Dropoff" strings (translated, see below).
- [x] Translations for new labels, "Pickup"/"Dropoff" strings and any other english-hardcoded string we could have.
- [ ] Display the loading spinner when we assign the order from the topbar right button (same behaviour as when we un/assign from the dispatch tasklist screen).
- [x] Make sure it works from both dispatcher and courier sections.
- [x] Make sure all tests passes!


package.json
```diff
@@ -17,6 +17,9 @@
     "@babel/helper-get-function-arity": "^7.16.7",
     "@datadog/mobile-react-native": "^2.4.3",
     "@datadog/mobile-react-navigation": "^2.4.3",
+    "@gluestack-style/react": "^1.0.57",
+    "@gluestack-ui/config": "^1.1.20",
+    "@gluestack-ui/themed": "^1.1.73",
```
alexsegura, line 22: Gluestack is now at version 3, version 1 is abandoned. 
I also work on this, you can cherry pick the installation in #2029

src/providers/UIProvider.tsx
alexsegura, line 1: Nice idea, I will do the same 👍

src/navigation/order/Order.tsx
```diff
@@ -1,23 +1,21 @@
...
-import TaskMiniMap from '../task/components/MiniMap';
-import OrderDetails from './components/OrderDetails';
 import { createCurrentTaskList } from '../../shared/src/logistics/redux/taskListUtils';
 import { blueColor } from '../../styles/common';
+import Task from '../../types/task';
+import TaskMiniMap from '../task/components/MiniMap';
+import OrderAccordeon from './components/OrderAccordeon';
+import OrderDetail from './components/OrderDetail';
+import { RouteType } from './types';
+import { Box } from '../../../components/ui/box';
```
alexsegura, line 16: Why not use the "global" import like in the docs?

```
import { Box } from '@/components/ui/box';
```

The same applies to all other imports.

src/navigation/order/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+  );
+
+  return uniqueTags;
+};
+
+export function getTaskTitleForOrder(task: Task): string {
+  if (task.address?.name && task.address?.contactName) {
+    return `${task.address.name} - ${task.address.contactName}`;
+  }
+  if (task.address?.contactName) {
```
diegomanuel, line 106: Missing the `if (task.address?.name)` case!
You could do:
```
export function getTaskTitleForOrder(task: Task): string {
    const addrdata = []; (could be a `let` but lint might complain)
    if (task.address?.name) addrdata.push(task.address?.name);
    if (task.address?.contactName) addrdata.push(task.address?.contactName);
    return addrdata.join(" - ");
}
```

src/navigation/task/components/utils.ts
```diff
@@ -29,3 +29,19 @@ export const getName = task => {
     })
     .join(' ');
 };
+
+export const getPackagesSummary = task => {
```
diegomanuel, line 33: I think I already saw this function in another file I was reviewing :thinking: 
If that's the case, try to re-use one single function!

diegomanuel (approved):
Simply awesome work @warmijusti  :clap: 
Some final small adjustments, a few checks and there we go to finally merge it!
I'm already approving it thou.

src/styles/gluestack-theme.ts
alexsegura, line 1: Backporting `useColorModeValue` is ok, but most functions duplicate the logic in `src/styles/theme.ts`. 
Is it possible to move new functions there, and use `src/styles/theme.ts` everywhere for hooks?

diegomanuel:
Merged it with only the spinner thing missing but that's a minor detail.
The idea is to make this work available to @alexsegura so he can continue at https://github.com/coopcycle/coopcycle-app/pull/2029 :raised_hands:

Issue #213, opened 2019-09-02
Task notes are not sent


alexsegura:
Bug introduced in d8bdb3584b7c38a4680d9ac569a428cce222d56c

PR #2027, opened 2025-08-26
[PERTE-213] Generate recurrent orders from App

Gonzalosanto:
# Issue [#213](https://github.com/orgs/coopcycle/projects/18/views/1?pane=issue&itemId=96500408&issue=coopcycle%7Ccoopcycle%7C213)

Application sends a request from dispatch view to `api/recurrence_rules` to generate recurrent orders as web is doing.  

## Tasks
- [x] Send request to API from App
- [x] Make sure all tests passes

### DIFF
--- a/src/navigation/dispatch/components/GroupedTasks.tsx
+++ b/src/navigation/dispatch/components/GroupedTasks.tsx
@@ -58,7 +58,7 @@ export default function GroupedTasks({
   const date = useSelector(selectSelectedDate)
   const [generateOrders] = useRecurrenceRulesGenerateOrdersMutation()
 
-  useEffect(() => {generateOrders(date)}, [generateOrders, date]);
+  useEffect(() => {generateOrders(date.format('YYYY-MM-DD'))}, [generateOrders, date]);
 
   const unassignedTaskLists = createUnassignedTaskLists(unassignedTasks);
   // Combine unassigned tasks and task lists to use in SectionList


src/redux/api/slice.ts
```diff
@@ -173,7 +173,15 @@ export const apiSlice = createApi({
         body: patch,
       }),
     }),
-  }),
+    recurrenceRulesGenerateOrders: builder.mutation ({
+      query: date => ({
```
vladimir-8, line 177: I fixed a lot of warnings about passing non-serialisable data to redux when migrating to v2. I believe there should be a warning here as well. Is there? If there is we should pass a date as a string instead of a `moment`

diegomanuel: Done at https://github.com/coopcycle/coopcycle-app/pull/2027/commits/94a30c2237e435253bdb8336ed5efe76bbb98d03

src/navigation/dispatch/components/GroupedTasks.tsx
```diff
@@ -55,6 +55,10 @@ export default function GroupedTasks({
   const navigation = useNavigation();
   const tasksEntities = useSelector(selectTasksEntities);
   const allTaskLists = useSelector(selectTaskLists);
+  const date = useSelector(selectSelectedDate)
+  const [generateOrders] = useRecurrenceRulesGenerateOrdersMutation()
+
+  useEffect(() => {generateOrders(date)}, [generateOrders, date]);
```
diegomanuel, line 61: Now you need to do here: `generateOrders(date.format('YYYY-MM-DD'))`
Do the change and plz check the call is being made fine.

diegomanuel (changes requested):
Missing change, will not work otherwise!

src/navigation/dispatch/components/GroupedTasks.tsx
diegomanuel: Done at https://github.com/coopcycle/coopcycle-app/pull/2027/commits/7a4ed39c847d5c32ab1380c9451325fc581724a4

diegomanuel:
Detox tests are all failing but seems to be due to CI malfunction.
Locally, e2e tests passes just fine.
Merging this one then..!

PR #1981, opened 2025-04-16
Bump RN 0.79

r0xsh:
> [!NOTE]
> Since RN79 just got release and [expo will "skip"](https://expo.dev/changelog/react-native-78#why-will-there-be-no-dedicated-stable-expo-sdk-release-for-078) RN78. I may be worth upgrading to 79 instead of 78

## Android

- [ ] Fix `outlineWidth` cast from String issue, might be related to native-base being out of date
- [ ] Fix multiple `.length` on null, it seems text input are handled in a different way
- [ ] Fix react navigation modals
- [x] Fix sunmi lib: https://github.com/coopcycle/react-native-sunmi-printer-library/commit/6d0c0b8fdd4071cbe38d648001e3e5414bfe0f2b
  - [ ] Test it on a real sunmi
- [ ] Test expo related modules.
  - [ ] Camera
  - [ ] Map
  - [ ] Picker
  - [ ] Background-fetchs
    
## iOS

- [ ] Try to compile and see what's broken :shrug:

PR #2048, opened 2025-09-26
Update React Native to 0.79

alexsegura:
Supersedes #1981


r0xsh (commented):
Tested on Android on a decent phone ; The app is so slow it's borderline unusable :/
It's hard to pinpoint what can be the cause.
To be extra sure this slowness come from this branch, i tried the CoopCycle Beta build and was smooth.

Also since the new black buttons ; there is no feedback when a button is pressed adding a extra "bugyness" feeling to it

https://github.com/user-attachments/assets/ba374fbf-f9b2-4142-987a-f00dea81a083

vladimir-8:
> Tested on Android on a decent phone ; The app is so slow it's borderline unusable :/

It might be due to the redux debug middleware. See the warnings in the logs: `ImmutableStateInvariantMiddleware took 167ms, which is more than the warning threshold of 32ms. If your state or actions are very large, you may want to disable the middleware as it might cause too much of a slowdown in development mode. See https://redux-toolkit.js.org/api/getDefaultMiddleware for instructions. It is disabled in production builds, so you don't need to worry about that.`

@r0xsh could you check if it's slow in a production build variant?

PR #2076, opened 2025-11-19
[PERTE-581] Tasks unassigned when a new task is assigned from the mobile app

diegomanuel:
## Issue https://github.com/coopcycle/coopcycle/issues/581

I was trying to do a sanity check before updating any tasklist using the hook `useSetTaskListItems`, but since RTK always update the state, when I manually call the trigger function `refreshTaskLists` from:
```
const [refreshTaskLists] = apiSlice.endpoints.getTaskListsV2.useLazyQuery();
```
It causes problems after (for example) assigning a task: It sometimes duplicates the task in the dispatcher task lists screen:
https://github.com/user-attachments/assets/691555d5-9fe5-4950-abfd-6219581e7493

**That's why I ended up by avoiding to use RTK to fetch task lists and use directly:
`const httpClient = httpClientService.getClient();`**
By going "outside" of RTK, I'm not messing with the state and I can do the sanity check in a centralized function before trying to send changes to backend (just one new request done, no matter how many tasklists are going to be updated down below the flow).

### The hook in question is: `src/logistics/redux/hooks/useSetTaskListItems.js`

I added a `__taskListsSanityCheck` function that tries to check if all the `tasklist.items` from the state matches the ones received in the API call.
The ideal would be to work directly with the fresh data that comes from the API, not needing to pass `allTaskLists` and `tasksEntities` at `useSetTaskListItems({ allTaskLists, tasksEntities })` but that "raw data" from API needs some processing done at:
* `src/redux/logistics/taskListEntityReducers.ts`
  - The `itemIds` property by calling `taskListUtils.replaceItemsWithItemIds()`)
  - This "translation" from `items` to `itemsIds` makes not much sense to me.. but it's widely used, I didn't want to mess with it.
* `src/shared/src/logistics/redux/selectors.js`
  - The `tasksIds` property from `selectTaskLists = createSelector(...` at `newTaskList.tasksIds = _.uniq(orderedItems)`)
  - This property is used mainly for filtering purposes.


vladimir-8:
I've implemented a fix for https://github.com/coopcycle/coopcycle/issues/527 in https://github.com/coopcycle/coopcycle-app/pull/2092

Let's see if we get this issue again and apply this workaround if needed

